PR #9, opened 2021-04-20
Implement the "View Source" command to open files in a right-split editor

jansorg:
Closes https://github.com/applandinc/appmap-intellij-plugin/issues/1

Adds a callback handler for the AppMap JS application's "View Source" command. A Java/JS bridge is installed on the JCEF handler to register a callback in the JS environment.
Path lookup logic:
1. Lookup by relative path
2. If no match was found, query the index of filenames and return the file with a matching parent hierarchy.
3. If there's still no match an error message is displayed

@ptrdvrk
The newly created editor is not focused, the focus remains on the AppMap application. Is this correct?


ptrdvrk:
Hello @jansorg , the editor should receive focus when you open a source from an AppMap.

PR #19, opened 2021-04-28
Build with @appland/appmap version 2.3.4.

jansorg:
Closes https://github.com/applandinc/appmap-intellij-plugin/issues/18

@ptrdvrk This is using 2.3.4 now, I wasn't certain if that the right version to use.


ptrdvrk:
Great, thank you @jansorg . I've built the plugin locally with 2.3.4 and it looks good.

PR #28, opened 2021-05-05
Initial implementation of AppMap tool window

### LINKED ISSUE #26: AppMap mode V1 exists
[ ] AppMap panel with AppMap icon exists
[ ] AppMap panel lists all AppMaps in project folders alphabetically by their names
[ ] Clicking on an AppMap in the list opens the AppMap in a viewer or focuses on an existing viewer with the AppMap
[ ] AppMap list can be filtered by user-entered text, matching metadata / name of AppMaps; filter can be reset
[ ] AppMaps can be found with user-entered search phrase, AppMaps in the result set can be opened in a new/existing viewer

jansorg:
Closes #26 

- Adds an index containing the metadata of all `.appmap.json` files found in the current project
- Adds a toolwindow, which displays the list of all available AppMap files. For each item the name (i.e. `metadata/name`) and the filename is displayed
- Changes to `.appmap.json` files on disk trigger a refresh of the tree
- Double-clicking or ctrl+enter open the file in the AppMap editor (this follows IntelliJ's defaults)
- Typing when the tree is focused highlights the next matching item (as almost anywhere else in IntelliJ)

@ptrdvrk @dustinbyrne

Questions:
- The search filter is not yet implemented. My suggestion is a text input field at the top of the panel. Is that an alternative to the filter of VSCode, which is hidden in a submenu?

To do:
- [ ] The search filter
- [ ] Testing performance and UI refresh with many files
- [ ] Suppress tree updates when the panel is hidden to avoid UI freezes

Preview of the current status (due to a lack of test data the same appmap file was copied many times):
Screenshot_003

### DIFF
--- a/.run/intellij-plugin [runIde].run.xml
+++ b/.run/intellij-plugin [runIde].run.xml
@@ -1,0 +1,23 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="intellij-plugin [runIde]" type="GradleRunConfiguration" factoryName="Gradle">
+    <ExternalSystemSettings>
+      <option name="executionName" />
+      <option name="externalProjectPath" value="$PROJECT_DIR$" />
+      <option name="externalSystemIdString" value="GRADLE" />
+      <option name="scriptParameters" value="" />
+      <option name="taskDescriptions">
+        <list />
+      </option>
+      <option name="taskNames">
+        <list>
+          <option value="runIde" />
+        </list>
+      </option>
+      <option name="vmOptions" value="" />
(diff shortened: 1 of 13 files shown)


dustinbyrne:
Hi @jansorg, this is looking fantastic! Regarding the search filter, a text input field at the top of the panel sounds fine. I'd actually wanted to do something similar in VS Code.

ptrdvrk:
Regarding scalability testing - we have been testing the Python AppMap agent  with a simple-to-map application with over 3,000 tests that generates 3,000 AppMap files, that should suffice. I'll provide details in a side channel.

PR #42, opened 2021-06-14
Collapsible panels for user milestones

jansorg:
This is still in an experimental state to add collapsible panels for user milestones.

Current status (steps are non-functional):
Screenshot_050


jansorg:
This is outdated now and a similar component has already been added to main.

Issue #45, opened 2021-06-17
The search results in the list of AppMaps in the AppMaps tool window auto-refresh after a .5s delay.


ptrdvrk:
Currently, ENTER triggers update of the search results. It would be nice if the results auto-updated when user pauses typing in the search box.

PR #47, opened 2021-06-22
#45: auto-refresh tool window when filter text changes

jansorg:
Tree is refreshed 500ms after typing in the search text input stopped

Closes  https://github.com/applandinc/appmap-intellij-plugin/issues/45


ptrdvrk (approved):
Looks good, works well, approving.

PR #50, opened 2021-07-27
Fix compilation with 2021.2 rc

jansorg:
- Fix compilation with latest rc build. This only affect source-compatibility, but not not binary compatibility. As far as I can tell version 0.4.1 is working fine with 2021.2 rc.
- Update to latest gradle-intellij plugin


ptrdvrk (approved):
Looking good, thank you @jansorg .

PR #52, opened 2021-07-29
Simplified quickstart UI

jansorg:
- [x] Adds an action 'AppLand: Open Quickstart'. Should this be kept? A: yes.
- [x] Display same list of links as in VSCode. Visited links are displayed with a different icon
- [x] Quickstart step "Install AppMap Agent" using the same JS application as in VSCode
- [x] Auto-open the "Install AppMap Agent" page after the first project is opened after a new plugin installation (not upgrades)

We can leave the following ones for the next update:
- [ ] Provide different icon of "Install AppMap Agent"  when needed
- [ ] Implement "Open AppMaps"

Screenshot_090


src/main/java/appland/toolwindow/milestones/MilestoneActions.java
```diff
@@ -0,0 +1,52 @@
...
+import org.jetbrains.annotations.NotNull;
+
+class MilestoneActions {
+    private static final Logger LOG = Logger.getInstance("#appland.milestones");
+
+    static void openQuickstart(@NotNull Project project) {
+        UserMilestonesEditorProvider.openUserMilestones(project);
+    }
+
+    /*static void installAppMapAgent(@NotNull Project project) {
```
ptrdvrk, line 15: What is this code, @jansorg ?

jansorg: That code was part of the first milestone iteration before we decided to make a simplified version. I wanted to keep it, but we could also move it into a branch and remove it here.

dustinbyrne (approved):

PR #67, opened 2021-11-30
Experimental REST endpoint to open an AppMap file

jansorg:
Please note, that this is still in an experimental state.

Example URL:
http://localhost:63342/api/appland?uri=/Users/user/dir/file.appmap
http://localhost:63343/api/appland?uri=/Users/user/dir/file.appmap
Try a few more ports if the above isn't working for you

Existing URL handler of JetBrains IDEs for testing:
http://localhost:63342/api/about
http://localhost:63343/api/about

ToDo:
- [ ] Restrict to *.appmap files and the corresponding internal file type
- [ ] Restrict to requests with a state query parameter (?)
- [ ] If possible, restrict to a strict set of referrers, e.g. http://localhost and https://app.land/
- [ ] Support relative file paths, locate them in the current project (?)
- [ ] Validate the state query parameter and pass it to the AppLand app, which is launched for the opened file.
- [ ] Test and implement support for Windows. It's currently unclear which file paths AppLand would send on Windows

### DIFF
--- a/src/main/java/appland/rest/ApplandHttpRequestHandler.java
+++ b/src/main/java/appland/rest/ApplandHttpRequestHandler.java
@@ -1,0 +1,76 @@
+package appland.rest;
+
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.fileEditor.OpenFileDescriptor;
+import com.intellij.openapi.project.ProjectManager;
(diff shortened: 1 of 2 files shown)


jansorg:
Closing because this is outdated now.

PR #71, opened 2022-01-17
Display an error message to the user if the file is too large to be loaded

### LINKED ISSUE #70: A warning is displayed for files which are too large
IntelliJ loads a truncated file preview for large files (at 20 megabytes, by default).
Currently, the truncated text is passed to the AppMap JS application, which fails to load it due to a syntax error of the incomplete JSON.

The AppMap editor should display a message to the user instead of trying to load the truncated content.

jansorg:
Closes #70 
This PR adds the needed changes to display an error message for too large files instead of trying to load the content.
The file size limit is user-configurable as a global, application-wide value within the IDE and defaults to 20 megabytes. Only advanced users of the IDE would customize this.

Although it's technically possible to load the complete content instead of the truncated content, this isn't working well with the current implementation.
Currently, the file content is turned into a JS method call string taking the complete content as argument. For large files this (and the needed JSON escaping of the content) makes things memory hungry and slow.
A method call is needed because the JS application is living in a sandboxed environment, which doesn't have access to `file://` URLs.

Possible workaround, which needs more investigation and more time than this PR: 
Add a custom scheme or resource URL handler and then instruct the JS application to load a file with our own custom protocol. The initial investigation of this turned out to be difficult, because custom scheme handlers must be added to the global JBCEF application before creating the first client. This isn't entirely under own control.
Additionally, we'd have to be very careful to not allow file system access for non-appmap files residing on the local filesystem.

Screenshot_20220117_122835

### DIFF
--- a/src/main/java/appland/editor/AppMapFileEditor.java
+++ b/src/main/java/appland/editor/AppMapFileEditor.java
@@ -20,10 +20,13 @@ import com.intellij.openapi.util.ActionCallback;
 import com.intellij.openapi.util.Disposer;
 import com.intellij.openapi.util.UserDataHolderBase;
 import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.openapi.util.io.FileUtilRt;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.vfs.VirtualFileEvent;
 import com.intellij.openapi.vfs.VirtualFileListener;
+import com.intellij.ui.SimpleTextAttributes;
 import com.intellij.ui.components.JBLoadingPanel;
+import com.intellij.ui.components.JBPanelWithEmptyText;
 import com.intellij.ui.jcef.*;
 import com.intellij.util.ui.UIUtil;
 import org.apache.commons.lang.StringEscapeUtils;
@@ -56,6 +59,7 @@ public class AppMapFileEditor extends UserDataHolderBase implements FileEditor {
     private static final Logger LOG = Logger.getInstance("#appmap.editor");
     private static final int LOADING_KEY = 0;
     private static final int CONTENT_KEY = 1;
+    private static final int ERROR_KEY = 2;
     private static final String READY_MESSAGE_ID = "intellij-plugin-ready";
 
     private final Project project;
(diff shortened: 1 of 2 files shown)


apotterri:
> The file size limit is user-configurable as a global, application-wide value within the IDE and defaults to 20 megabytes. Only advanced users of the IDE would customize this.

Is the file size limit really user-configurable? I don't see any way to do it....

apotterri (changes requested):
If the file-size limit is configurable, this PR needs some documentation describing how to change it. If it's not configurable, then things are fine as-is.

Also, how do updated versions get deployed to the JetBrains marketplace? Was Petr doing that?

jansorg:
@apotterri The file size is configurable, but only in a very technical way. See https://www.jetbrains.com/help/idea/tuning-the-ide.html#common-platform-properties for instructions. I wouldn't advertise this to users because it's something 99% of users won't do and it also side-effects for every other file in the IDE.
Petr was uploading new plugins to https://plugins.jetbrains.com/plugin/16701-appmap, based on the instructions at https://github.com/applandinc/appmap-intellij-plugin#releases. Please let me know if you need help with it.

You're already member https://plugins.jetbrains.com/organizations/AppLand, if you don't have upload permissions yet the other members should be able to grant you the needed permissions.

apotterri:
> @apotterri The file size is configurable, but only in a very technical way. See https://www.jetbrains.com/help/idea/tuning-the-ide.html#common-platform-properties for instructions. I wouldn't advertise this to users because it's something 99% of users won't do and it also has side-effects for every other file in the IDE.
> 
> Petr was uploading new plugins to https://plugins.jetbrains.com/plugin/16701-appmap, based on the instructions at https://github.com/applandinc/appmap-intellij-plugin#releases. Please let me know if you need help with it.
> 
> You're already member https://plugins.jetbrains.com/organizations/AppLand, if you don't have upload permissions yet the other members should be able to grant you the needed permissions.

Thanks for the info about configuring the file size. I agree that we don't need to tell users about this, unless they ask.

I don't see any way for me to upload a new version of the plugin, so I'm assuming I don't yet have permission to do it. I'll work with others to get me going.

It looks to me like we should be able to automate upload of new versions, using the Gradle plugin: https://plugins.jetbrains.com/docs/intellij/deployment.html . Do you know of any reason this wouldn't work?

apotterri (approved):

jansorg:
@apotterri Sure, automatic uploads are possible, e.g. with GitHub actions. I offered to work on CI, but this task never made it onto the lists of things to work on.

apotterri:
Great, thanks for confirming that.

PR #74, opened 2022-02-11
Update @appland/components and @appland/diagrams to the latest available versions

jansorg:
This PR updates the following dependencies and also contains the packaged JS bundle at `appland/dist/`. 

Please note, that this also adds `appland/dist/entry.bundle.js.LICENSE.txt`, which was generated by npm based on the dependencies. This license file will be part of the distributed plugin ZIP file.

```json
    "@appland/components": "^1.23.0",
    "@appland/diagrams": "^1.5.1",
```

@apotterri I'm not very familiar with the JS application, but think that this is including the filtering we discussed. I'd be glad if you could test this in the IDE and confirm that it's working as expected.

Screenshot_20220211_174101


jansorg:
@apotterri I've updated the version to 0.7.7 and also updated the changelog

apotterri (approved):

PR #82, opened 2022-08-18
CI with Travis

jansorg:
Adds a `.travis.yml` file to enable CI using the default Gradle integration.
It fixes a few things in the Gradle build setup to build successfully on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+dist: jammy
+
+language: java
+jdk: openjdk17
+
+# https://github.com/travis-ci/travis-ci/issues/1147#issuecomment-441393807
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,19 @@
+dist: jammy
+
+language: java
+jdk: openjdk17
```
apotterri, line 4: Would it be useful to test against other Java versions as well?

apotterri (approved):

.travis.yml
jansorg: This is only used to compile and not to run the sandbox, which is using it's own JetBrains JDK. The source level is set to Java 11 already to let the compiler check the sources.

PR #95, opened 2022-08-31
Install guide: Web application

jansorg:
This adds the JS web application part of the install guide feature.
The existing folder "appland-user-milestones" will be removed when the install guide feature is activated in the UI.


dustinbyrne (approved):
Approving, but I have a couple of comments/questions which you may or may not want to address now:

- `appland-install-guide/dist` is not gitignored - should it be?
- There are now three JS projects in the repo, can they be merged down to one?
- There's a mix of `npm` and `yarn`, intended?

Also, it seems like in this PR there's no means of actually loading this frontend, is that right? This is purely an additive change.

jansorg:
Thanks!
- the `dist` folder is currently build locally once and committed into Git. I hope to make this part of the Gradle build process to avoid adding `dist` to version control.
- The old `appland-user-milestones` will be removed in another PR, so it's two JS applications (AppMap viewer and install guide). Unless there's a benefit to merge those two, I'd leave it at that
- Unfortunately caused by the history of the development, will be cleaned up

PR #105, opened 2022-09-07
fix: Add external service objects

dustinbyrne:
Without this, AppMaps containing external service calls will fail to be indexed.

### DIFF
--- a/src/main/java/appland/index/ClassMapItemType.java
+++ b/src/main/java/appland/index/ClassMapItemType.java
@@ -10,11 +10,12 @@ public enum ClassMapItemType {
     Folder(1, "folder"),
     Database(2, "database"),
     Package(3, "package", "/"),
+    ExternalService(4, "external-service"),
     // children types
-    Route(4, "route"),
-    Query(5, "query"),
-    Class(6, "class", "::"),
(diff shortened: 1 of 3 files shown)


src/main/java/appland/index/ClassMapItemType.java
```diff
@@ -10,11 +10,12 @@ public enum ClassMapItemType {
...
+    ExternalService(4, "external-service"),
     // children types
-    Route(4, "route"),
-    Query(5, "query"),
-    Class(6, "class", "::"),
-    Function(7, "function", "#");
+    Route(5, "route"),
+    Query(6, "query"),
+    Class(7, "class", "::"),
+    Function(8, "function", "#");
```
dustinbyrne, line 18: I believe this will break existing indexes (?). Should the `ExternalService` constant instead be incremental instead of shifting all the existing constants?

jansorg: That's correct, but method `getVersion()` is used to force a rebuild after changes to the indexes. Just increase the version in `src/main/java/appland/index/IndexUtil.java` .

jansorg:
I updated the index version before merging this

PR #126, opened 2022-09-16
feat: Add telemetry

### LINKED ISSUE #86: IntelliJ extension includes telemetry

dustinbyrne:
This change adds some basic telemetry capabilities and includes a few telemetry events

Resolves #86

### DIFF
--- a/src/main/java/appland/notifications/AppMapNotifications.java
+++ b/src/main/java/appland/notifications/AppMapNotifications.java
@@ -17,8 +17,11 @@ import javax.swing.*;
 
 import static appland.AppMapBundle.lazy;
 
+import java.util.function.Consumer;
+
 public final class AppMapNotifications {
     public static final String REMOTE_RECORDING_ID = "appmap.remoteRecording";
+    public static final String TELEMETRY_ID = "appmap.telemetry";
 
     public static void showExpiringRecordingNotification(@NotNull Project project,
                                                          @Nullable String title,
@@ -86,5 +89,27 @@ public final class AppMapNotifications {
 
             notification.notify(project);
         }, ModalityState.any());
+    }
+
+    public static void showTelemetryNotification(@NotNull Project project,
+    @Nullable String title,
+    @NotNull String content,
+    @NotNull NotificationType type,
+    @NotNull Consumer<Boolean> onDismiss) {
+        ApplicationManager.getApplication().invokeLater(() -> {
+            Notification notification = new AppMapFullContentNotification(
+                        TELEMETRY_ID, null,
+                        title, null, content,
+                        type, null
+                );
+
+            var denyAction = NotificationAction.create(lazy("telemetry.permission.deny"), (e, n) -> {
+                onDismiss.accept(false);
+                n.expire();
+            });
+
+            notification = notification.addAction(denyAction);
+            notification.notify(project);
+        });
     }
 }
--- a/src/main/java/appland/startup/AppLandStartupActivity.java
+++ b/src/main/java/appland/startup/AppLandStartupActivity.java
@@ -3,6 +3,8 @@ package appland.startup;
 import appland.installGuide.InstallGuideViewPage;
 import appland.installGuide.InstallGuideEditorProvider;
 import appland.settings.AppMapApplicationSettingsService;
+import appland.telemetry.TelemetryService;
+
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.startup.StartupActivity;
@@ -17,6 +19,8 @@ public class AppLandStartupActivity implements StartupActivity {
             AppMapApplicationSettingsService.getInstance().setFirstStart(false);
 
             openToolWindowAndQuickstart(project);
+
+            TelemetryService.getInstance().notifyTelemetryUsage(project);
         }
     }
 
--- a/src/main/java/appland/startup/AppLandStartupActivity.java
+++ b/src/main/java/appland/startup/AppLandStartupActivity.java
@@ -20,7 +20,9 @@ public class AppLandStartupActivity implements StartupActivity {
 
             openToolWindowAndQuickstart(project);
 
-            TelemetryService.getInstance().notifyTelemetryUsage(project);
+            var telemetry = TelemetryService.getInstance();
+            telemetry.notifyTelemetryUsage(project);
+            telemetry.sendEvent("plugin:install");
         }
     }
 
--- a/src/main/java/appland/telemetry/AppInsightsClient.java
+++ b/src/main/java/appland/telemetry/AppInsightsClient.java
@@ -1,0 +1,61 @@
+package appland.telemetry;
+
+import com.intellij.openapi.application.ApplicationInfo;
+import com.intellij.openapi.diagnostic.Logger;
+import com.intellij.util.Urls;
+import com.intellij.util.io.HttpRequests;
+
+import appland.AppMapPlugin;
+import appland.telemetry.data.Tag;
+import appland.telemetry.data.TelemetryEvent;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+import org.jetbrains.annotations.NotNull;
+
+
+/*
+ * Based on:
+ * https://github.com/Azure/azure-sdk-for-java/blob/4a53adf6274ced5af8243983042b5e32bac85bd7/sdk/
+ * monitor/azure-monitor-opentelemetry-exporter/src/main/java/com/azure/monitor/opentelemetry/
+ * exporter/AzureMonitorExporterBuilder.java
+ */
+class AppInsightsClient {
+    private final @NotNull String ingestionEndpoint;
+    private final @NotNull Gson gson = new GsonBuilder().create();
+
+    AppInsightsClient(@NotNull String ingestionEndpoint) {
+        this.ingestionEndpoint = ingestionEndpoint;
+    }
+
+    public void track(@NotNull TelemetryEvent event) {
+        var userId = Identity.getMachineId();
+        var osVersion = System.getProperty("os.version");
+
+        event.property("common.os", System.getProperty("os.name"))
+            .property("common.platformversion", System.getProperty("os.version"))
+            .property("common.jvmversion", System.getProperty("java.version"))
+            .property("common.extversion", AppMapPlugin.getDescriptor().getVersion())
+            .property("common.intellijversion", ApplicationInfo.getInstance().getFullVersion())
+            .property("common.product", ApplicationInfo.getInstance().getBuild().getProductCode())
+            .property("common.source", "JetBrains")
+            .tag(Tag.OsVersion, osVersion)
+            .tag(Tag.UserId, userId)
+            .tag(Tag.SessionId, Session.getId());
+
+        try {
+            var url = Urls.newFromEncoded(ingestionEndpoint).resolve("/v2/track").toExternalForm();
+            var json = gson.toJson(event);
+
+            HttpRequests.post(url, "application/json")
+                .tuner(t -> t.setRequestProperty("Accept", "application/json"))
+                .connect(req -> {
+                    req.write(json);
+                    return req;
+                });
+        } catch (Exception e) {
+            Logger.getInstance(getClass()).warn(e);
+        }
+    }
+}
--- a/src/main/java/appland/telemetry/AppMapRecordListener.java
+++ b/src/main/java/appland/telemetry/AppMapRecordListener.java
@@ -1,0 +1,32 @@
+package appland.telemetry;
+
+import org.jetbrains.annotations.NotNull;
+
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.project.Project;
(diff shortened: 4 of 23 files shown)


src/main/java/appland/telemetry/AppInsightsClient.java
```diff
@@ -0,0 +1,56 @@
...
+            var json = gson.toJson(event);
+            System.out.println(json);
+            HttpRequests.post(url, "application/json")
+                .tuner(t -> t.setRequestProperty("Accept", "application/json"))
+                .connect(req -> {
+                    req.write(json);
+                    return req;
+                });
+        } catch (Exception e) {
+            e.printStackTrace();
```
jansorg, line 53: Better use `Logger.getInstance(getClass()).warning(e);` or `Logger.getInstance(getClass()).error(e);` to log this

src/main/java/appland/telemetry/AppMapRecordListener.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.startup.StartupActivity;
+
+import appland.files.AppMapFileChangeListener;
+
+public class AppMapRecordListener implements Disposable, StartupActivity {
```
jansorg, line 12: It's better to register this as a listener inside `<applicationListeners>` inside `plugin.xml`:
```xml
    <applicationListeners>
        ...
        <listener topic="appland.files.AppMapFileChangeListener"
                  class="appland.telemetry.AppMapRecordListener"/>
    </applicationListeners>

```

with this implementation:

```java
package appland.telemetry;

import appland.files.AppMapFileChangeListener;
import org.jetbrains.annotations.NotNull;

import java.util.concurrent.atomic.AtomicBoolean;

public class AppMapRecordListener implements AppMapFileChangeListener {
    private final @NotNull AtomicBoolean appMapRecordedThisSession = new AtomicBoolean(false);

    @Override
    public void afterAppMapFileChange() {
        if (appMapRecordedThisSession.compareAndSet(false, true)) {
            TelemetryService.getInstance().sendEvent("appmap:record");
        }
    }
}
```

`AtomicBoolean` (or a `volatile boolean`) is preferred over a `Boolean` to avoid potential threading issues, e.g. by concurrent updates of the field.

Please note that it's possible that `afterAppMapFileChange` is called after a deletion of an `.appmap.json` file.

src/main/java/appland/telemetry/Identity.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.jetbrains.annotations.NotNull;
+
+import com.intellij.ide.util.PropertiesComponent;
+
+public class Identity {
+  @NotNull protected final static String MachineIdKey = "appland.machine_id";
```
jansorg, line 17: nit: `private`?

src/main/java/appland/telemetry/Identity.java
```diff
@@ -0,0 +1,78 @@
...
+  @NotNull protected final static Stream<String> IgnoredMacAddresses =
+      Stream.of("00:00:00:00:00:00", "ff:ff:ff:ff:ff:ff", "ac:de:48:00:11:22");
```
jansorg, line 19: Streams a usually only used a temporary objects. At least I never to it :)

You could make this a set instead to simplify the lookup:
```
@NotNull private final static Set<String> IgnoredMacAddresses =
      Set.of("00:00:00:00:00:00", "ff:ff:ff:ff:ff:ff", "ac:de:48:00:11:22");
// ...
       if (IgnoredMacAddresses.contains(mac)) {
          continue;
        }
```

src/main/java/appland/telemetry/data/BaseData.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class BaseData {
+    @SerializedName("ver")
+    @NotNull final Integer version = 2;
+
+    @SerializedName("name")
+    @NotNull String name;
+
+    @SerializedName("properties")
+    @Nullable HashMap<String, String> properties;
```
jansorg, line 18: nit: It's better to use an interface for the properties instead, `Map<String,Map>`. This makes it more generic and simplifies later changes.

src/main/java/appland/telemetry/data/TelemetryEvent.java
```diff
@@ -0,0 +1,47 @@
...
+    @NotNull final String name = "Microsoft.ApplicationInsights.Event";
+
+    @SerializedName("time")
+    @NotNull String time;
+
+    @SerializedName("iKey")
+    @NotNull String iKey;
+
+    @SerializedName("tags")
+    @Nullable HashMap<String, String> tags;
```
jansorg, line 22: nit: `Map<String, String>`

src/main/java/appland/telemetry/TelemetryService.java
```diff
@@ -0,0 +1,67 @@
...
+    public static @NotNull TelemetryService getInstance() {
+        return ApplicationManager.getApplication().getService(TelemetryService.class);
+    }
+
+    public interface TelemetryEventBuilder {
+        BaseData build(@NotNull BaseData eventData);
+    }
+
+    public boolean isEnabled() {
+        return AppMapApplicationSettingsService.getInstance().isTelemetryEnabled();
```
jansorg, line 30: Do you have to check the env variable `APPMAP_TELEMETRY_DISABLED` here?

src/main/java/appland/startup/AppLandStartupActivity.java
```diff
@@ -17,6 +19,10 @@ public void runActivity(@NotNull Project project) {
             AppMapApplicationSettingsService.getInstance().setFirstStart(false);
 
             openToolWindowAndQuickstart(project);
+
+            var telemetry = TelemetryService.getInstance();
+            telemetry.notifyTelemetryUsage(project);
```
jansorg, line 24: Please note that this notification is sometime only shown in the log (toolwindow "Notifications".) Only two or three notifications are shown at once and there are sometimes more than that at startup. If AppMap's notifications happens to be in the set of hidden notifications, it would only be listed in the "notifications" panel and not shown again.

jansorg (changes requested):
Great work! Definitely an improvement to use HTTP to send telemetry and does the right background job handling when sending requests to avoid blocking the UI.
I made a few comments and suggestions. Please let me know when you need more details!

I think you need to rebase on master, a few imports were unresolved but things worked fine after I rebased.

src/main/java/appland/telemetry/TelemetryService.java
dustinbyrne: A native setting is be preferred here. In the future we can use this flag to determine whether or not `APPMAP_TELEMETRY_DISABLED` is passed to child processes. However, none of the commands we're using at the moment will even emit telemetry so this will be a no op.

src/main/java/appland/telemetry/AppInsightsClient.java
```diff
@@ -0,0 +1,56 @@
...
+    public void track(@NotNull TelemetryEvent event) {
+        var userId = Identity.getMachineId();
+        var osVersion = System.getProperty("os.version");
+
+        event.property("common.os", System.getProperty("os.name"))
+            .property("common.platformversion", System.getProperty("os.version"))
+            .property("common.jvmversion", System.getProperty("java.version"))
+            .property("common.extversion", AppMapPlugin.getDescriptor().getVersion())
+            .property("common.intellijversion", ApplicationInfo.getInstance().getFullVersion())
+            .property("common.source", "IntelliJ")
```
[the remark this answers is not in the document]
dustinbyrne: Yes this sounds better, thank you.

src/main/java/appland/startup/AppLandStartupActivity.java
dustinbyrne: As long as it meets JetBrains standards for collecting telemetry (which it seems like it does?) that's fine by me.

src/main/java/appland/telemetry/AppMapRecordListener.java
dustinbyrne: Thanks! I've switched the startup activity to a listener.

> Please note that it's possible that `afterAppMapFileChange` is called after a deletion of an `.appmap.json` file.

In regard to this, I had to change the  `AppMapAsyncFileListener` interface a bit to inform the listener about the types of changes that have occurred. This felt a bit better than creating a new type of listener that notifies only on create or update. Let me know what you think.

src/main/java/appland/startup/AppLandStartupActivity.java
jansorg: As far as I know only paid plugin have to show an opt-in notification. Other plugins don't have to follow this rule strictly, so the AppMap plugin is fine, AFAIK.

jansorg: We could add a setting to the UI settings of the AppMap plugin

src/main/java/appland/telemetry/AppInsightsClient.java
```diff
@@ -0,0 +1,61 @@
...
+            .property("common.extversion", AppMapPlugin.getDescriptor().getVersion())
+            .property("common.intellijversion", ApplicationInfo.getInstance().getFullVersion())
+            .property("common.product", ApplicationInfo.getInstance().getBuild().getProductCode())
+            .property("common.source", "JetBrains")
+            .tag(Tag.OsVersion, osVersion)
+            .tag(Tag.UserId, userId)
+            .tag(Tag.SessionId, Session.getId());
+
+        try {
+            var url = Urls.newFromEncoded(ingestionEndpoint).resolve("%s/v2/track").toExternalForm();
```
jansorg, line 48: `resolve("%s/v2/track")` should be `resolve("/v2/track")` instead. `%s` was used for the host, which is now passed into `newFromEncoded`.

jansorg (changes requested):
Thanks, this is looking great! Just one minor change, feel free to merge after that.
I like the change of the listener method, it's definitely useful.

src/main/java/appland/startup/AppLandStartupActivity.java
dustinbyrne: I'm not sure - this does appear to speak generally to any plugin on the marketplace:
https://plugins.jetbrains.com/legal/approval-guidelines

src/main/java/appland/telemetry/AppInsightsClient.java
dustinbyrne: Whoops. I meant to take that out! Thanks for the catch :+1:

src/main/java/appland/startup/AppLandStartupActivity.java
jansorg: Oh, you're right, my statement was wrong.

We could
- make this opt-in instead of opt-out
- show the notification again at startup until the user either accepted or rejected the telemetry
- add a setting to the AppMap settings UI to allow to turn it off/on again later
What do you think?

src/main/resources/META-INF/plugin.xml
```diff
@@ -85,6 +88,9 @@
                            key="notification.remoteRecording.displayName"/>
         <statusBarWidgetFactory implementation="appland.remote.RemoteRecordingStatusBarFactory"/>
 
+        <notificationGroup id="appmap.telemetry" displayType="BALLOON"
```
jansorg, line 91: if you change this to `<notificationGroup id="appmap.telemetry" displayType="STICKY_BALLOON"` (i.e. "STICKY_BALLOON"), then the notification won't expire automatically after 10 second.
I think that's better for important things like telemetry.

src/main/java/appland/notifications/AppMapNotifications.java
```diff
@@ -87,4 +90,29 @@ public static void showExpandedRecordingNotification(@NotNull Project project,
...
+                );
+
+            var denyAction = NotificationAction.create(lazy("telemetry.permission.deny"), (e, n) -> {
+                onDismiss.accept(false);
+                n.expire();
+            });
+
+            notification = notification.addAction(denyAction);
+            notification.whenExpired(() -> {
+                onDismiss.accept(true);
```
jansorg, line 113: This is always enabling telemetry, even after the user clicked "Opt out".
A click on "Opt out" first sets it to disabled and then expires the notification. The expiry triggers this `whenExpired`, which resets it back to default.

jansorg (changes requested):
See below for notes on a possible bug.

It might be an alternative to show the notification until the user either selected "opt out" or "accept". 
The `enableTelemetry` field could become a nullable `Boolean`, e.g. 
```
    @Getter
    @Setter
    private volatile @Nullable Boolean enableTelemetry = null;
```
`null` would indicate "no choice", `true` would be "opt in" and `false` "opt out".

src/main/java/appland/notifications/AppMapNotifications.java
dustinbyrne: Thanks, I'm just going to drop this statement.

jansorg (approved):

PR #129, opened 2022-09-19
Add AppMap executor to extend Java run configurations with the AppMap agent

### LINKED ISSUE #137: Java "run configurations" are modified to enable map recording

jansorg:
- [x] IntelliJ's own Java run configurations
- [x] Gradle run configurations
- [x] Maven run configurations
- [x] Integration tests with run configurations, if possible
- [x] Mapping to the build system's output directory
- [ ] Icon for the new toolbar action

Demonstration with a unmodified clone of spring-petclinic. Here, just one test is executed.
https://www.loom.com/share/204a5cbf88204dbaba47abb320beaef8

I had to change the Gradle build structure of the plugin, because we have to add optional dependencies for the Java, Gradle, and Maven IDE plugins. Non-optional dependencies would requires Java, Gradle, Maven everywhere, which would exclude all of RubyMine, PhpStorm, etc.

This PR contains an integration test for the IntelliJ java build and execution. Unfortunately, I don't see a good way to add integration tests for the Gradle and Maven integrations. AFAIK the test support is not part of the public SDK and replicating this is a lot of work with high maintenance costs.

Error when an incompatible JDK was detected:
image

Fixes #137 .


apotterri:
@jansorg @brikelly 

After watching the Loom and changing the project's JDK, I was able to run this end-to-end. I used it to generate AppMaps from a clean spring-petclinic clone. That was *awesome*!

"After watching the Loom and changing the project's JDK" points to a couple of ways I think it could be improved:

* "Start with AppMap" is really buried in that popup. I first looked for an AppMap run config here:
image

Does it make sense to add it?

* Is it possible to detect the version of the JDK currently set for the project? If we can determine it's unsupported, can we prompt the user to change it? My project was configured to use 17, and it failed (of course). Had I been a new user, I probably would have abandoned at that point.

I think both of those issues are significant enough that I'd like to see them addressed before we release this. I'm not sure it even make sense to merge this PR without doing so.

apotterri:
Oh, I noticed this broken tooltip, too: 
image

jansorg:
> * "Start with AppMap" is really buried in that popup. I first looked for an AppMap run config here:
> Does it make sense to add it?

@apotterri Could you post a screenshot of the popup you mentioned?

We're not adding new run configurations, but a new way to execute existing types of run configurations.
I agree that this may be a bit confusing to new users. Think of it as the difference between "Run" and "Debug" - same settings, but different way to execute and use it.

There are several ways to execute them:
1. By the popup of line markers next to test methods (as in the Loom video)
2. By clicking the AppMap logo in the toolbar (right of the run configuration drop down)
   Screenshot_20221007_203455
3. For example, in the context menu of Maven tasks (there are several more like this)
    Screenshot_20221007_203636

Does this make sense?

> * Is it possible to detect the version of the JDK currently set for the project? If we can determine it's unsupported, can we prompt the user to change it? My project was configured to use 17, and it failed (of course). Had I been a new user, I probably would have abandoned at that point.

Yes, that's good point. It should be possible to detect it, although it's a bit difficult to get right. Each module of a project may have it's own JDK settings and a run configuration may reference one or more modules.

jansorg:
> > By clicking the AppMap logo in the toolbar (right of the run configuration drop down)
> > The logo is disabled when I first open the project:
> as are the rest of the icons. What's the minimum that needs to happen to get them enabled?

These buttons are enabled as soon as there's a run configuration selected in the box on the left which says "Add configuration...". "Add configuration" is displayed when there's no configuration at all, i.e. in a new project.

> The problem I see here is that, when opening a project for the first time, I don't know what to do. It looks like this: 
> and it's not at all obvious that I need to click down to find a test, then click on one of the line markers next to get to that popup.

That's the default of JetBrains IDEs. We can't do much about this, unfortunately. As you see in the screenshot, only the first two icons (Run and Debug) are shown in the context menu. The remaining icons (Coverage and AppMap) are moved into the submenu of the context menu. I don't think we should attempt to push Run or Debug into the submenu instead of AppMap, because Run and Debug are the most commonly used options and IMO user's won't like this kind of override.

The alternative is to create a new Java run configuration (Application, JAR, Gradle, Maven) via "Add configuration". The context menu is just a shortcut to do the same as "Add configuration" for the current file or Java test method.

JetBrains' documentation is at https://www.jetbrains.com/help/idea/running-applications.html

> > Yes, that's good point. It should be possible to detect it, although it's a bit difficult to get right. Each module of a project may have it's own JDK settings and a run configuration may reference one or more modules.
> > Can we put off the detection until the user actually picks the menu item? That seems like a point where we'd know exactly what's about to happen.

Yes, that's the best time for the check. We could provide feedback to the user that another JDK is needed. 
But even then the difficulty is to find the JDK, which is actually used by the run configuration. I'll see what I can do :)

jansorg:
> Oh, I noticed this broken tooltip, too: image

This is fixed now

jansorg:
This is ready for review now

apotterri (approved):
Looks great! Thanks very much.

PR #134, opened 2022-09-20
Migration from appland.com to appmap.io and update Github URL

petecheslock:
We have setup permanent redirects for our domain on appland.com -> appmap.io.


petecheslock:
@jansorg I'm not sure how to confirm those large diffs - or test them.  They are basically from the output of search/replace from "appland.com" to "appmap.io" so they _should_ be safe.

jansorg:
@petecheslock You can safely ignore all files under `src/test/data`. These files are only used for our tests. Do you have an easy way to revert just the changes inside this dir? I think that would make this PR a bit easier to read.

jansorg (changes requested):
I think most of this is fine.
Many changes are in the bundled JS sources. Updating to a new version of the JS component, which already have the new URLs would be better, I think.
I don't trust search and replace inside minified sources...

If the new version also carries other, incompatible changes, then we may have to go with search-and-replace.

Reverting changes inside `src/test/data` would make this PR a bit smaller. It's only generated data.

petecheslock:
I can go and pull out the changes within those directories to make this more manageable.

petecheslock:
ok - @jansorg i've pulled out changes to files in `src/test/data` which makes the PR much simpler.

petecheslock:
Also here is the PR i created for the appmap-js project.  https://github.com/getappmap/appmap-js/pull/740

jansorg:
Unless this PR is urgent, I'd prefer to wait for https://github.com/getappmap/appmap-js/pull/740, because that more reliable than search&replace in minified sources.

petecheslock:
@jansorg it's merged now should be good to go.

jansorg (approved):
Seems fine to me. I briefly tested the webview JS applications and couldn't find an issue with the URLs.
I'd prefer to just link to updates JS applications via `package.json`, but that's not possible without introduction possible issues, because only newer versions are available with the updated URLs. Let's merge now and update the JS applications in new PRs.

I pushed a commit to fix two failing tests.

PR #145, opened 2022-10-12
Fix `verifyPlugin`

jansorg:
Reverts the commit which removed `verifyPlugin` and fixes the gradle setup to only execute `verifyPlugin` for the main module, which is published.


apotterri (changes requested):
Does this change really need to generate a new GitHub/Marketplace release? If not, please change the commit type from `fix:` to `ci:`

jansorg:
@apotterri Thanks, I've updated the commit

apotterri (approved):

PR #157, opened 2022-10-25
fix: Emit appmap:create events

dustinbyrne:
All the file events were coming through as refresh, so no notifications containing create/modify events would ever fire. `AppMapAsyncFileListener` had appeared to be unused due to this.

### DIFF
--- a/plugin-core/src/main/java/appland/files/AppMapAsyncFileListener.java
+++ b/plugin-core/src/main/java/appland/files/AppMapAsyncFileListener.java
@@ -12,8 +12,6 @@ import java.util.stream.Collectors;
 
 /**
  * Async file listener to notify about changes to .appmap.json files.
- * We don't react to changes from a file system refresh.
- * {@link  VirtualFileManagerLister} is taking care of refresh events.
  */
 @SuppressWarnings("UnstableApiUsage")
 public class AppMapAsyncFileListener implements AsyncFileListener {
@@ -40,10 +38,6 @@ public class AppMapAsyncFileListener implements AsyncFileListener {
     }
 
     private static boolean isAppMapFileChange(@NotNull VFileEvent event) {
-        if (event.isFromRefresh()) {
-            return false;
-        }
-
         // only treat renames as valid property changes
         if (event instanceof VFilePropertyChangeEvent) {
             if (!((VFilePropertyChangeEvent) event).isRename()) {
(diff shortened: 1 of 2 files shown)


plugin-core/src/main/java/appland/files/AppMapAsyncFileListener.java
```diff
@@ -40,17 +40,6 @@ public class AppMapAsyncFileListener implements AsyncFileListener {
     }
 
     private static boolean isAppMapFileChange(@NotNull VFileEvent event) {
-        if (event.isFromRefresh()) {
```
jansorg, line 43: The intention was to let `appland.files.VirtualFileManagerLister` notify after a filesystem refresh finished and this listener to handle non-refresh events. But this split approach is not suitable if you need to know about e.g. new files picked up by a refresh.

I'd remove the note of this class's documentation about the now removed refresh handling.

plugin-core/src/main/java/appland/files/AppMapAsyncFileListener.java
```diff
@@ -40,17 +40,6 @@ public class AppMapAsyncFileListener implements AsyncFileListener {
...
-            return false;
-        }
-
-        // only treat renames as valid property changes
```
jansorg, line 47: Is there a reason why you removed this?
A property change event could also be "encoding change" and others. With this code removed, these would be send as `Other` changes.

plugin-core/src/main/java/appland/files/AppMapAsyncFileListener.java
jansorg: This change could lead to multiple notifications about refresh events, e.g. once for a new file from a refresh and again by `VirtualFileManagerLister` after the refresh finished.
I checked the consumers of the notifications, they're debouncing so this change shouldn't make it flicker, for example.

jansorg (approved):

jansorg:
@dustinbyrne can this be merged?

dustinbyrne:
Let me circle back on this. I'd actually thought this was waiting on changes given the feedback.

jansorg:
I only expected possible changes by you to address feedback, if necessary. Please let me know if I should make changes…

plugin-core/src/main/java/appland/files/AppMapAsyncFileListener.java
dustinbyrne: Looks like I may have missed the negation. I'll add it back.

plugin-core/src/main/java/appland/files/AppMapAsyncFileListener.java
dustinbyrne: In practice I was only observing new files picked up by refresh. I'm not clear on the conditions under which a file event which did not originate from a refresh would occur.

dustinbyrne:
I've updated the PR. Let me know if this is sufficient.

jansorg:
@dustinbyrne I think this is ready for merging...

PR #165, opened 2022-11-17
fix: Use `/authn_provider` for authentication

dustinbyrne:
`/authn_provider/localhost` has been deprecated and will be removed in the future.


jansorg:
Is this still supporting the redirect to localhost URLs?

dustinbyrne:
Yes. This is a single route we'll use for both JetBrains and VS Code moving forward.

dustinbyrne:
It's only accepting redirect URLs to `http://localhost` or `vscode://appland.appmap` at the moment

PR #170, opened 2022-11-21
Pre-release tweaks: upgrade the AppMap agent and specify an authentication source during login

dustinbyrne:
This change:
- Bumps the bundled AppMap agent version to latest
- Adds a parameter to the authentication request indicating the source of the authentication (`JetBrains`). This is used by the backend to determine the description of a users API key.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,7 +8,7 @@ ideVersionVerifier=IC-2021.3,IC-2022.1.3
 
 lombokVersion=1.18.24
 
-appMapJvmAgent=1.15.1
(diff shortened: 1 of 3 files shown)


jansorg (approved):
Looks fine to me!
I think you also need to update the supported JDK version at https://github.com/getappmap/appmap-intellij-plugin/blob/dd5f7c7a67dca77f67ebd0a4323ca760ecadedb3/plugin-core/src/main/java/appland/execution/AppMapJvmExecutor.java#L30

I'm not sure what JDK version the latest agent is supporting now...

PR #176, opened 2022-12-08
Bundle appmap-java agent v1.15.4

jansorg:
Can be tested by launching the IDE with the plugin and running a run configuration with the AppMap executor.
I verified that the bundled jar is the new version.


apotterri (approved):
Looks good, thanks for this.

kgilpin:
Note that releases may be disabled right now because of the Travis
credentials leak issue

PR #201, opened 2023-02-06
Telemetry to better understand sign in

### LINKED ISSUE #198: Telemetry exists to better understand user signin behavior

ahtrotta:
Fixes #198 

Adds telemetry events to help understand where to put the sign-in CTA.

**Telemetry events added**

- [x] Click on a Docs link from the tree view (`appland.appmap/click_docs_link_in_tree`)
- [x] Use the search bar in the AppMap sidebar (`appland.appmap/sidebar_search_focused`)
- [x] Click on a list item in the AppMap sidebar (`appland.appmap/select_object_in_sidebar`)
- [x] Click on the `Dependency Map` tab in the AppMap (`appland.appmap/click_tab` with Properties['appmap.click_tab.tabId'] == 'viewComponent')
- [x] Click on the `Trace View` tab in the AppMap (`appland.appmap/click_tab` with Properties['appmap.click_tab.tabId'] == 'viewFlow')
- [x] Click on the Filter button in the AppMap header (`appland.appmap/click_filter_button`)
- [x] Click on the Refresh button in the AppMap header (`appland.appmap/reset_diagram`)


plugin-core/src/main/java/appland/webviews/appMap/AppMapFileEditor.java
```diff
@@ -151,6 +153,38 @@ public void setState(@NotNull FileEditorState state) {
...
+            case "sidebarSearchFocused":
+                telemetryService.sendEvent("sidebar_search_focused");
+                return null;
+
+            case "clickFilterButton":
+                telemetryService.sendEvent("click_filter_button");
+                return null;
+
+            case "clickTab":
+                if (message == null) return null;
```
jansorg, line 165: nit: nesting `if` statements may be a bit easier to read, similar to the handler of `viewSource`. But that's mostly personal choice.

appland/webview.js
```diff
@@ -38,8 +38,13 @@ export function mountWebview() {
 
   messages.on('init', ({data: initData}) => {
     // messages emitted by the webview
-    app.$on('viewSource', (location) => vscode.postMessage({command: 'view-source', location}))
+    app.$on('viewSource', (location) => vscode.postMessage({command: 'viewSource', location}))
     app.$on('uploadAppmap', () => vscode.postMessage({command: 'uploadAppMap'}))
+    app.$on('sidebarSearchFocused', () => vscode.postMessage({command: 'sidebarSearchFocused'}))
```
jansorg, line 43: The alternative to the telemetry-related messages would be a message like `telemetryMessage` with the properties embedded. But that would move telemetry handling into the JS source...
If there's going to be a lot more messages like these, then I'd probably move it here to avoid the verbose handling of telemetry.

plugin-core/src/main/java/appland/webviews/appMap/AppMapFileEditor.java
jansorg: because this is in a `switch` and it's always returning `null`, this would things a bit easier to follow.

jansorg (approved):
Looks good to me! 
Only a few minor notes, feel free to ignore because it's mostly personal choice.

Issue #127, opened 2022-09-17
Code Objects view should exist


brikelly:
This should match what the VS Code extension supports, with a section of the panel dedicated to users browsing the catalog of indexed objects found in all the maps:

![Image](https://user-images.githubusercontent.com/573108/190864239-70498865-fca9-41a8-860f-16f0a573e6ec.png)

ahtrotta:
@jansorg here is a spec: https://docs.google.com/document/d/14kVc7C-7ZSITcVEk6skmXthLtNyR0lxVRrL_bc-EL1U/edit

Please let me know if you need anything isn't clear!

PR #210, opened 2023-02-18
Implement code objects

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/127

Implements code objects, similar to VSCode.
It adds a new panel to the AppMap tool window to list the available code objects. It displays the same three root node as the VSCode plugin.
It's extending the index of ClassMap to provide the needed data.

- [x] Make Code objects a collapsible panel
- [x] Verify why VSCode attempts to navigate to the source (.java) file for package/function nodes
- [x] Open code source files for the package view
- [x] Compare tree rendering with VSCode
- [x] Fix refresh after changes to AppMap files and reindex
- [x] Fix sort order of HTTP request items
- [ ] Open the function/query/route in the AppMap
- [ ] Should the panel be shown when a user is not authenticated?
- [ ] If you (i.e. AppMap) would like better icons, then you should provide them

## Status
Packages:
image

HTTP requests:
image

Queries:
image

Navigation popup if there are multiple AppMaps for a node:
image


ahtrotta:
When I click on a leaf in the `Code` tree in VS Code, the source file where the function is defined is opened:

image

But in JetBrains it shows a list of AppMaps and then opens the test file:

image

Could you have the behavior in JetBrains mimic VS Code?

ahtrotta:
Also, when the name of a SQL Query is long the text overlaps and is illegible:

image

In VS Code, the text is truncated and then there's a hover state where the whole query is legible:

image

I'm not sure what the options for addressing this are in JetBrains, but let me/Brian know how you think this can be addressed. Thanks!

jansorg:
@ahtrotta I've fixed the issues you found. The latest commits update the navigation for function items and the truncation of the displayed label.

ahtrotta (approved):
Looks good!

PR #224, opened 2023-03-16
fix: don't generate a default config

### LINKED ISSUE #223: The plugin no longer generates a default config
Once https://github.com/getappmap/appmap-java/pull/170 gets merged, the Java agent will start to generate a default config. This functionality shouldn't be duplicated in the plugin.

apotterri:
**This should not be released until the Java agent with [these changes](https://github.com/getappmap/appmap-java/pull/170) gets incorporated.**

The Java agent now generates a default config. Don't duplicate that functionality in the plugin.

Fixes #223 .


kgilpin:
Isn't it the case that IntelliJ can generate an AppMap config with package
names, but the agent cannot?

dividedmind (commented):
I also have some doubts if this should be removed. Seems like this, by leveraging the editor's code model, can do substantially better at generating the config file than the simplistic approach used in the agent.

kgilpin:
We'll keep this for now - not least because we don't know how long it will take for the appmap-java agent changes to fully roll out.

apotterri:
> I also have some doubts if this should be removed. Seems like this, by leveraging the editor's code model, can do substantially better at generating the config file than the simplistic approach used in the agent.

In practice, projects mostly seem to conform to the convention that their source will be in `src/main/java`. So, while using IntelliJ's code model is better, it's probably not substantially better.


> We'll keep this for now - not least because we don't know how long it will take for the appmap-java agent changes to fully roll out.

The plugin bundles the version of `appmap-java` it requires, so this wouldn't have been a concern.

I've opened https://github.com/getappmap/appmap-intellij-plugin/pull/226 to update the plugin so it also sets `shallow: true`, to match the new functionality in `appmap-java`.

PR #226, opened 2023-03-18
fix: only record top-level packages by default

apotterri:
Set "shallow: true" in the generated config, to limit the number of method calls captured by default.


kgilpin:
Ok, let’s try this.

kgilpin (commented):
I think maybe we should hold this back until we see how effective the new Stats are.

dividedmind (commented):
I'm also not sure if this is warranted. I would definitely like to see more discussion and analysis before I approve it.

kgilpin:
Yes. Now that we have large AppMap support with integrated stats I think we
can safely configure for full recording by default.

apotterri:
Ok. I'll close this, and also undo this change: https://github.com/getappmap/appmap-java/commit/9ff949b68081e79329006693cdd6633097fcf9ce#diff-c4944a78aa4deb0b304f48cd9441fbf970694d9373fe0053bb40ba8651fbe8ebR231

PR #227, opened 2023-03-21
Support sequence diagrams

jansorg:
Updates the AppMap JS webview components to the latest versions to support AppMap sequence diagrams. 
It also adds support for the new "Export to SVG" button.

This is best tested manually, because there are no UI tests for the webview integration.

UI to choose the location of the SVG:
image

Warning if the file already exists:
image

Default editor view of the new SVG file:
image


ahtrotta (approved):
It looks good and works as expected. 

I have two comments, but we might want to merge and release this and then address my comments later, because they're minor and we want to release sequence diagrams in JetBrains as soon as possible.

1. The default filename for the exported SVG file is `appMap.svg`. I think it would be better if the default filename was the name of the appMap being exported.

3. When the exported SVG is large, it's not very useful to display it in JetBrains. I can't seem to zoom in so I can't read anything:

image

Issue #233, opened 2023-03-30
All Code Objects should link to AppMaps


brikelly:
The code objects list in JetBrains should always open an AppMap rather than jump to code. Currently, HTTP requests and Database queries do correctly open AppMaps (or prompt if there are multiple) but code functions go to code. It should change to instead always open an AppMap.

That would make it consistent with our VS Code extension as well. Plus, Java users in IntelliJ already have a package/class explorer so we don't want to duplicate that.

PR #234, opened 2023-04-04
Open AppMap for all types of code objects

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/233

Removes the special handling of function code objects. These code objects are now opened in an AppMap, i.e. all code objects are now treated in the same way.

### DIFF
--- a/plugin-core/src/main/java/appland/toolwindow/codeObjects/ClassMapItemNavigatable.java
+++ b/plugin-core/src/main/java/appland/toolwindow/codeObjects/ClassMapItemNavigatable.java
@@ -6,6 +6,7 @@ import appland.index.AppMapMetadata;
 import appland.index.AppMapMetadataIndex;
 import appland.index.ClassMapItemType;
 import appland.index.ClassMapTypeIndex;
+import appland.webviews.appMap.AppMapFileEditorState;
 import com.intellij.openapi.application.ReadAction;
 import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.progress.ProgressManager;
@@ -68,7 +69,7 @@ public class ClassMapItemNavigatable implements Navigatable {
             case 0:
                 return;
             case 1:
-                FileEditorManager.getInstance(project).openFile(filesToMetadata.get(0).first, requestFocus);
+                openCodeObjectEditor(project, requestFocus, filesToMetadata.get(0));
                 return;
             default:
                 JBPopupFactory.getInstance()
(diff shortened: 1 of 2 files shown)


ahtrotta (changes requested):
I apologize that I didn't make this clear earlier, but we want to open the AppMap to the relevant code object when a user clicks on a code object in the menu. 

In VS Code, after selecting a map, the map is opened to the relevant location.

image


In JetBrains, we want to do something similar. The current change in this PR does everything correctly except open the AppMap to the relevant code object.

The way this can be implemented is by using the `setState` method in the appmap Vue component [here](https://github.com/getappmap/appmap-js/blob/main/packages/components/src/pages/VsCodeExtension.vue#L979). In VS Code, we send a series of messages back and forth from VS Code to the frontend during the loading process, including a message with the desired state for the appmap.

After the map has finished loading, we send the state using the `setState` method [here](https://github.com/getappmap/vscode-appland/blob/master/web/src/appmapView.js#L47).  The `setState` method expects a JSON object with a `currentView` and a `selectedObject`. The `currentView` should be `viewComponent` and the `selectedObject` should be the `fqid` of the code object. For example:

```
{
  "currentView": "viewComponent",
  "selectedObject": "function:actionpack/ActionController::Instrumentation#process_action"
}
```

Let me know if there's anything that's not clear and I'll make a more detailed spec for you.

jansorg:
@ahtrotta Thanks! I added the setState in https://github.com/getappmap/appmap-intellij-plugin/pull/234/commits/a7e6cfb360e23fafb4284f50d521a329d4b15729.
This preselects the method in the left pane of the AppMap view, but does not seem to highlight the node as in your screenshot.

I tried to test this VSCode with the latest version of AppMap, but it's still opening the source file and not the AppMap view.

ahtrotta:
I just ran this branch locally (`./gradlew runIde`) and I noticed that there aren't any sequence diagrams. What version of `@appland/components` does this branch use? The selection not occurring in the map might be because the version of `@appland/components` is old. 

image

jansorg:
@ahtrotta I just rebased this branch on `main` and also updated to the latest appmap/components. Please try again.

I still think that there's something wrong with the navigation, but I'm not sure.

ahtrotta:
In VS Code, it opens the source file and there's an AppMap pin next to the function. If you hover over the pin it will give you the option to `Open code object in AppMap`. In JetBrains, we want to mimic the `Open code object in AppMap` functionality directly.

image

ahtrotta:
Yeah, I see the `Sequence Diagram` tab but there is still a problem with selecting the current code object in the map.

jansorg:
I rebased onto main and think that it's now working in the same way as the VSCode plugin.
At last for https://github.com/land-of-apps/sample_app_6th_ed/ the AppMap view highlights the chosen method.
It does not work for projects like spring-petclinic, but that's also the case for VSCode.

ahtrotta:
Awesome! It seems to open the map with the correct function selected!

I've found a bug, though. If I open [this repo](https://github.com/land-of-apps/sample_app_6th_ed/tree/codespaces-devcontainer), generate maps, and then navigate to the code objects tree view (`Code --> ActionController --> Instrumentation --> process_action`):

image

If I select a map (for example `Site_layout layout links`):

image

Then I reset the map view with the `Clear selection` button:

image

If I close the map and then I re-open the map from the `AppMaps` list:

image

The map will open with the same function selected again, even though I cleared that selection in the previous step:

image

So it seems like the selection state is being passed into the map again even though the map is being opened from the `AppMaps` list and not the `Code Objets` tree view.

ahtrotta (approved):
This change looks good! It makes sense to tackle the bug mentioned at the end of the conversation as a separate issue.

Issue #235, opened 2023-04-06
Sequence Diagrams are the default view when maps are opened


brikelly:
Currently the default is Dependency Map, but we wish to change that to sequence diagrams. This was recently done in VS Code here: https://github.com/getappmap/vscode-appland/pull/662

PR #236, opened 2023-04-10
Show sequence diagram view by default

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/235

Updates to the latest AppMap component and adds the property to show the sequence diagram view by default.


ahtrotta (approved):
Looks good, thanks!

Issue #232, opened 2023-03-28
JetBrains requires sign-in for usage


brikelly:
@ahtrotta to add details on what the UX should be in order to match the VS Code flow.

ahtrotta:
Here is the [spec](https://docs.google.com/document/d/1KGnRJ3LaOeekNCn6-Ia2IoudBQVr4PtEwIPZ9A_Ba7k/edit?usp=sharing).

Please let me know if anything is unclear.

ahtrotta:
@brikelly, are we putting this on hold?

brikelly:
No, @Daniel-Warner-X is creating a new design for the page, but we still need to build the plumbing to render it. So for now we'll proceed to get this prepared, and drop in the UI changes from Dan once they're ready.

jansorg:
@ahtrotta Is this component already available via NPM? https://github.com/getappmap/appmap-js/commit/5c18782dccae16a61a121aefa199d87697448d27#diff-8f0db08fec54128a50fad8a9e12c6afc7451d85716eac2a988afda5d16345dd5R1

ahtrotta:
Yes, that should be available

jansorg:
Thanks! I was able to import it after updating to the latest appmap/components.

PR #237, opened 2023-04-10
Show SignIn webview in toolwindow if a user is not yet authenticated

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/232

- Displays a new webview in the AppMap tool window if the user is not yet signed in. It's adding a new webview in directory "appland-signin"
- Replaces the webview with the usual AppMap content after the user signed in to the AppMap server
- Sends telemetry for webview event "click-sign-in-link", it should be following VSCode's implementation

Open issues:
- [x] The black background of the webview panel isn't covering all available space. Is this the responsibility of the Vue component or of my HTML wrapper?

image


jansorg:
@ahtrotta https://github.com/getappmap/appmap-intellij-plugin/pull/237/commits/0ffa8c3f1d446a95bc84dc62636adf290b3cd042 fixes the background. 
I wasn't able to reproduce your rendering bug. Are you still seeing this with the latest commit? Is moving the scrollbar fixing or updating the rendering glitches?
There's not much we could do to fix the webkit rendering, though. Which OS and IDE version did you use for testing?

ahtrotta (approved):
The rendering bug I was seeing has been fixed with the latest update, so everything looks good!

PR #257, opened 2023-04-25
Handle "clear selection" of AppMap webview

### LINKED ISSUE #246: Reset AppMap editor state
I've found a bug, though. If I open [this repo](https://github.com/land-of-apps/sample_app_6th_ed/tree/codespaces-devcontainer), generate maps, and then navigate to the code objects tree view (`Code --> ActionController --> Instrumentation --> process_action`):

image

If I select a map (for example `Site_layout layout links`):

image

Then I reset the map view with the `Clear selection` button:

image

If I close the map and then I re-open the map from the `AppMaps` list:

image

The map will open with the same function selected again, even though I cleared that selection in the previous step:

image

So it seems like the selection state is being passed into the map again even though the map is being opened from the `AppMaps` list and not the `Code Objets` tree view.

_Originally posted by @ahtrotta in https://github.com/getappmap/appmap-intellij-plugin/issues/234#issuecomment-1511612926_

jansorg:
Closes #246 

This PR adds support for "clear selection" to properly reset the editor state when an AppMap is reopened or restored at startup by the IDE.


jansorg:
@ahtrotta A more generic fix might be to just support "stateChanged". VSCode only seems to use it [for the current selection](https://github.com/getappmap/vscode-appland/blob/e21f4a127133be669e0038ab4149183ef140d9d7/web/src/appmapView.js#L141). What's the expected behavior of "stateChanged"?

ahtrotta:
`stateChanged` is emitted from the front end when either of the following are changed in the map view:

* The current view (i.e. `Dependency Map`, `Sequence Diagram`, or `Trace View`)
* The selected object

In VS Code we are listening for this emitted event just to send telemetry when the selected object is changed.

jansorg:
@ahtrotta This PR is ready for review now. It doesn't add handling of "stateChanged" to follow VSCode's implementation. If this event should be handled for telemetry, could you open a new issue, please?

ahtrotta:
This works better, but the behavior is still a little strange. If I open an appmap using the `Code Objects` tree view, it correctly opens the the map with the function selected:

image

If I then select another node:

image

And then close the appmap and re-open it using the `AppMaps` selector, it opens back to the selected function from when I opened it using the `Code Objects` view:

image

I would have expected this to either:
* Open the appmap without a state (as if I had just clicked the reset button), or
* Open the appmap to the last state that it was in (with the `Renderers` class selected)

So, can you either save the state on exit and then re-open to the same state, or just erase the state when the appmap is closed?

jansorg:
@ahtrotta https://github.com/getappmap/appmap-intellij-plugin/pull/257/commits/abfd0e596c95a7926805e73958d0572ca9b2dc95 removes the restoring of the editor state.
Previously, the state was only restored when an AppMap editor in the current session. Reopening the project or restarting the IDE cleared the IDE's internal memory of editor state.

This PR now keeps the webview state independent of the IDE's editor state.

ahtrotta (approved):
It's working great now!

PR #260, opened 2023-05-06
Fix support for large AppMaps on Windows

jansorg:
Fixes non-working handling of large AppMaps on Windows. The "stats" command was printing ANSI escape codes to STDOUT, which were breaking JSON parsing. This only happened on Windows, though. Running CI on Windows would have found this problem.

The other commits fix CI on Windows and should also fix the flaky CI with Travis. Background activity, which is often delayed on slow systems, was breaking the tests.


ahtrotta (approved):
I tested it and it works great, thanks Joachim!

Issue #266, opened 2023-05-12
java.lang.ExceptionInInitializerError


covailab:
Name | Value
----:|:-----
Plugin|appland.appmap
Version|0.34.1
IDE|AI-222.4459.24.2221.9971841
OS|Windows 10 10.0
Last action|

### Stacktrace
```text
java.lang.ExceptionInInitializerError
	at appland.toolwindow.signInView.SignInViewPanel.<init>(SignInViewPanel.java:32)
	at appland.toolwindow.AppMapToolWindowFactory.createContentPanel(AppMapToolWindowFactory.java:71)
	at appland.toolwindow.AppMapToolWindowFactory.updateToolWindowContent(AppMapToolWindowFactory.java:45)
	at appland.toolwindow.AppMapToolWindowFactory.createToolWindowContent(AppMapToolWindowFactory.java:28)
	at com.intellij.openapi.wm.impl.ToolWindowImpl.createContentIfNeeded$intellij_platform_ide_impl(ToolWindowImpl.kt:533)
	at com.intellij.openapi.wm.impl.ToolWindowImpl.scheduleContentInitializationIfNeeded$intellij_platform_ide_impl(ToolWindowImpl.kt:512)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.doShowWindow(ToolWindowManagerImpl.kt:923)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.showToolWindowImpl(ToolWindowManagerImpl.kt:863)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.activateToolWindow$intellij_platform_ide_impl(ToolWindowManagerImpl.kt:597)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.activateToolWindow$intellij_platform_ide_impl$default(ToolWindowManagerImpl.kt:576)
	at com.intellij.openapi.wm.impl.ToolWindowManagerImpl.activated$intellij_platform_ide_impl(ToolWindowManagerImpl.kt:1833)
	at com.intellij.toolWindow.StripeButton$1.actionPerformed(StripeButton.kt:61)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2313)
[...]

PR #280, opened 2023-05-31
Warn about missing support for JCEF webviews 

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/266

This PR displays a panel with an error message instead of attempting to use a JCEF webview on IDEs, where it's unsupported. For example, Android Studio does not support it by default.

I'd be glad for thoughts on the error message and if we should link to a webpage with further details on this or not.

image

### DIFF
--- a/plugin-core/src/main/java/appland/toolwindow/AppMapToolWindowFactory.java
+++ b/plugin-core/src/main/java/appland/toolwindow/AppMapToolWindowFactory.java
@@ -12,6 +12,7 @@ import com.intellij.openapi.wm.ToolWindowFactory;
 import com.intellij.openapi.wm.ToolWindowManager;
 import com.intellij.openapi.wm.ex.ToolWindowManagerListener;
 import com.intellij.ui.content.ContentFactory;
+import com.intellij.ui.jcef.JBCefApp;
 import org.jetbrains.annotations.NotNull;
 
 import javax.swing.*;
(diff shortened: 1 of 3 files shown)


ahtrotta:
This looks good, but I think it would be helpful if the message displayed to the user included a link to a site that explained how they can fix the problem. [This article](https://joachimschuster.de/posts/android-studio-fix-markdown-plugin/) seems a little outdated, but linking to a similar article about how to enabled JCEF support would help the user to fix the problem themselves.

jansorg:
I agree to add a link, but I'd prefer to only link to a webpage of JetBrains or AppLand. I'm not aware of a similar page by JetBrains, though. Would it be possible to add a page on AppLand's own website?

ahtrotta:
For the error message you can link to [this article](https://intellij-support.jetbrains.com/hc/en-us/articles/206544879-Selecting-the-JDK-version-the-IDE-will-run-under).

```
Your IDE does not support JCEF HTML webviews.
Please change the Boot Java Runtime for the IDE by following [these directions].
```

jansorg:
@ahtrotta Thanks, that's helpful! I missed that link.
Here's the updated panel. Unfortunately, there's no support for inline-links, only for complete lines styled like a link. Also, there's no automatic line wrapping, but the lines are hopefully short enough.
image

ahtrotta (approved):
Great! Looks good to me. Let's merge it once the build passes.

Issue #276, opened 2023-05-30
Index of AppMap metadata index should not read .appmap.json files


jansorg:
Reading and indexing .appmap.json may have a major impact on indexing performance. 
Instead, we should rely on the meta .json files created by the AppMap CLI tools to retrieve this metadata. 

This would also fix the problem, that AppMap files are shown in IntelliJ, which would not be shown in VSCode.

Location of metadata:
> name - available in metadata.json
request count - available in canonical.httpServerRequests.json
query count - available in canonical.sqlNormalized.json
function call count - not available, but number of functions (rather than function call events) in the AppMap is available via the classMap.json

PR #281, opened 2023-05-31
Index AppMap metadata files instead of .appmap.json files

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/276


jansorg:
@ahtrotta Is there anything left for me to do? I wasn't sure why you assigned this issue to me.

ahtrotta:
I just manually tested this branch and the indexing is not working well for me. I opened a fresh clone of `spring-petclinic` and generated some appmaps using the tests. I can see that the appmaps were indexed because the expected folders/files are present, but I'm not seeing any appmaps in the AppMap panel:

image

If I close IntelliJ and then re-open it, the appmaps briefly appeared but then they disappear again.

jansorg:
@ahtrotta Thanks! You're right, it didn't work. 
I've pushed https://github.com/getappmap/appmap-intellij-plugin/pull/281/commits/63a4c0e14bb978be3499e248151b23f8f0c0c4a6 to fix this. The cause of the problem was, that AppMap data in excluded parent folders were not properly indexex by the IDE. Our tests did not catch this, because the indexing is different in test mode (and we can't override this, AFAIK).

PR #282, opened 2023-05-31
fix: Update documentation links

dividedmind:


jansorg (approved):
:tada: for the sed expression :)

PR #290, opened 2023-06-07
Use new AppMap scopes where possible

jansorg:
#275 added methods to declare AppMap search scopes. `main` was changed in the meantime and this PR updates the recently added scopes to use the new AppMap scopes instead.

### DIFF
--- a/plugin-core/src/main/java/appland/actions/OpenRecentAppMapAction.java
+++ b/plugin-core/src/main/java/appland/actions/OpenRecentAppMapAction.java
@@ -1,6 +1,7 @@
 package appland.actions;
 
 import appland.AppMapBundle;
+import appland.index.AppMapSearchScopes;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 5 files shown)


plugin-core/src/main/java/appland/actions/OpenRecentAppMapAction.java
```diff
@@ -1,20 +1,20 @@
...
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.search.FilenameIndex;
-import com.intellij.psi.search.ProjectScope;
 import org.jetbrains.annotations.NotNull;
 
 /**
  * Locates the most recently modified .appmap.json file in the project and opens it.
  * This needs an index and thus can't run when DumbMode is active.
- *
+ * <p>
```
ahtrotta, line 17: Was this intentional?

ahtrotta (approved):

plugin-core/src/main/java/appland/actions/OpenRecentAppMapAction.java
jansorg: This was added by reformatting the file, so it‘s not strictly needed but would be added again the next time the code is formatted.

Issue #289, opened 2023-06-06
IntelliJ remote recordings are being saved to a non-indexed location


brikelly:
To reproduce: 
* Use Spring-Petclinic or similar to run an application with AppMap enabled
* Start a remote recording
* Stop the remote recording
* Accept the default storage location and save the map
* The map opens in the viewer, but Code Objects will not get updated, nor will Analysis show new findings

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/573108/7935f6df-c279-4d2f-830a-152042ce3a0b)

The issue is that `appmap.yml` specifies that the `appmap_dir` is `target/appmap`, but that setting is not being honored. The storage location dialog box defaults to a different location, `tmp/appmap`. This causes the indexer not to see the new map.

It looks like test recordings are going to the correct location:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/573108/d7b4e425-d7b6-4bc3-a216-f54ecd966a74)

As part of fixing this issue, please check any other locations which might hardcode `tmp/appmap` or otherwise not honor the `appmap_dir` setting in `appmap.yml`.

PR #293, opened 2023-06-07
Use better default of storage location of "Stop AppMap recording"

jansorg:
closes https://github.com/getappmap/appmap-intellij-plugin/issues/289

- Drop the file path input from the dialog of the "Stop Remote Recording" action
- Default to `target/appmap` instead of `tmp/appmap`

The storage location of "Stop Remote Recording" now looks into appmap.yml and then defaults to  `target/appmap` if no `appmap_dir` setting was found.


ahtrotta (approved):
Looks good and worked when I tried it, thanks!

Issue #267, opened 2023-05-16
Maps from sibling folders show up in the JetBrains map list


brikelly:
@kgilpin can reproduce this, and reports that when he has a project open in folder A, he can also see maps from sibling folder B. This could cause significant problems for the plugin as it tries to watch / index too many files, and also create a confusing UX.

Unknown how to reliably reproduce this on other systems yet.

Maybe there is an `appmap.yml` in a parent folder outside the workspace. But even in that case, `appmap_dir` should be respected. Generally speaking, only valid appmap_dir should be considered when listing or indexing AppMaps.

jansorg:
> Maybe there is an `appmap.yml` in a parent folder outside the workspace. But even in that case, `appmap_dir` should be respected. Generally speaking, only valid appmap_dir should be considered when listing or indexing AppMaps.

 @brikelly Could you clarify the expcted behavior? Currently, only AppMap files found in the current project's directories are shown. But these files are currently not filtered by the `appmap_dir` values of appmap.yml config files.

dividedmind:
@ahtrotta you moved this to *In Review* but there is no pull request attached?

jansorg:
@dividedmind Sorry, my bad. I added https://github.com/getappmap/appmap-intellij-plugin/pull/295

PR #295, opened 2023-06-08
Don't show AppMaps, which are not contained in the current project

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/267

This PR removes the use of the `everythingScope()`, because it's likely that this caused the unrelated AppMaps to show up.
The documentation of the everything scope is very limited and unclear in the SDK. Even though it takes a `Project` to create it, it's still accepting files outside that project.


ahtrotta:
@jansorg could you rebase this with main before I review it? Thanks!

jansorg:
@ahtrotta Rebased. This is ready for review now.

ahtrotta (approved):

Issue #288, opened 2023-06-06
PyCharm installer mangles the console output


brikelly:
When installing within a Python project (I used Misago) in PyCharm, the console used for the installer CLI has messy and unreadable text breaks:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/573108/cdd11dd4-6d67-45c6-865c-0fc841aa5433)

brikelly:
@jansorg this issue from June looks identical to the issue we discovered recently. Do you have any idea why it would have regressed?

jansorg:
@brikelly could you link the other issue, please?
This issue is about garbled output, the only other issue I know was about a json parsing error in the pycharm console.

brikelly:
The other issue is this one: https://github.com/getappmap/appmap-js/issues/1308

PR #248, opened 2023-04-18
Prune large AppMaps using the CLI

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/244

This is best tested with a project containing large and giant AppMaps.

- Executes the AppMap CLI to purge large AppMaps
- Always adds the "stats" property (loaded via AppMap CLI) to the JSON passed to the AppMap webview
- Adds tests to test the CLI with the prune and stats commands

PR #296, opened 2023-06-08
Execute the AppMap install command with a PTY

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/288

The install command needs a PTY, e.g. to query for user input on the terminal.
This is a regression of https://github.com/getappmap/appmap-intellij-plugin/pull/248. The prune command did not work with a  PTY. This PR adds a test and fix to ensure that the install command is executed with a PTY terminal.


ahtrotta (approved):
Awesome, thank you!

Issue #287, opened 2023-06-06
The latest Java agent jar is downloaded upon every plugin startup


brikelly:
Currently the plugin bundles the Java agent jar within the plugin zip. This means that it will progressively go out of date for users who install the plugin once but never update it. 

Instead, we would like the plugin to always download the latest Jar from the official location (if a newer one exists) upon each plugin startup.

The plugin should be added to the user's home directory under a `.appmap/lib/java` sub-folder on Linux and Mac, and the equivalent on Windows. 

The agent jar should also be saved without version numbers to predictably easy to find by our various tools and code. The name should be `appmap.jar`. Therefore the full path of where the jar should be saved would be `~/.appmap/lib/java/appmap.jar`. It should be saved with a versioned file name, and symlinked as `appmap.jar` to prevent changes to existing projects or run configurations. If symlinking fails (e.g. on Windows) then we should ensure that the latest jar file gets saved as `appmap.jar` regardless.

brikelly:
@jansorg yes, but my comments in the issue above should address that (by placing it into the user's home directory), no?

kgilpin:
I think you can find JAR version in the Manifest. Then you can easily determine if a new version of the JAR needs to be fetched

PR #300, opened 2023-06-13
Download the latest AppMap Java agent at startup

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/287

With this PR, the Java agent JAR file is downloaded at startup.
The download is only made if the current IDE provides Java support (i.e. IntelliJ IDEA Ultimate or Community). The agent file is only needed for Java-based run configurations. Limiting the download to Java projects is unreliable IMHO, because projects can contain mixed of any kind of language and Java support may be added after a project is opened.

This follows the implementation of VSCode, as outlined by @ahtrotta :
> 1. If the latest JAR file version already exists on disk, then do nothing and return.
> 2. Look for a file in the target directory called appmap-X.X.X.jar.downloading. If this file already exists, and it's less than 5 minutes old, return and do nothing - because another VSCode instance is already performing the update.
> 3. Touch/create the lock file appmap-X.X.X.downloading. Set mtime to now if it already exists.
> 4. Download the new JAR file.
> 5. Delete the lock file.
> 6. Symlink or copy the downloaded JAR file to appmap.jar

### DIFF
--- a/plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
+++ b/plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
@@ -80,18 +80,18 @@ public final class AppMapJavaAgentDownloadService {
             return false;
         }
 
-        // e.g. appmap-1.7.2.downloading
-        var lockFilePath = assetDownloadPath.resolveSibling(latestAsset.getFileName().replace(".jar", ".downloading"));
+        // e.g. appmap-1.7.2.jar.downloading
+        var lockFilePath = assetDownloadPath.resolveSibling(latestAsset.getFileName() + ".downloading");
         assert !assetDownloadPath.equals(lockFilePath);
 
         if (!guardWithLockFile(lockFilePath, () -> downloadAgentJarFile(indicator, latestAsset, assetDownloadPath))) {
             return false;
         }
 
-        // create a symbolic link pointing to "appmap.jar"
+        // create a relative symbolic appmap.jar pointing to the downloaded JAR file
         try {
             Files.deleteIfExists(agentFilePath);
-            Files.createSymbolicLink(agentFilePath, assetDownloadPath);
+            Files.createSymbolicLink(agentFilePath, agentFilePath.getParent().relativize(assetDownloadPath));
         } catch (IOException e) {
             // copy the downloaded file to "agent.jar" as a fallback,
             // e.g. on system where symbolic links are unsupported
@@ -161,7 +161,7 @@ public final class AppMapJavaAgentDownloadService {
      * @return The path of the directory, where the AppMap Java agent should be stored (~/.appmap/lib/java).
      */
     @Nullable Path getOrCreateAgentDir() {
-        var agentDirPath = Paths.get(System.getProperty("user.dir")).resolve(Paths.get("appmap", "lib", "java"));
+        var agentDirPath = Paths.get(System.getProperty("user.dir")).resolve(Paths.get(".appmap", "lib", "java"));
         try {
             Files.createDirectories(agentDirPath);
             return agentDirPath;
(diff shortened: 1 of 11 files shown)


ahtrotta:
Is this ready for review? I see that you tagged me for review, but the build is failing and it is still in the `In Progress` column on the board.

jansorg:
Sorry, that was my mistake. This PR is ready for review now.

plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
```diff
@@ -0,0 +1,173 @@
...
+    private @Nullable Path getJavaAgentPath() {
+        var agentDir = getOrCreateAgentDir();
+        return agentDir == null ? null : agentDir.resolve(AGENT_LINK_FILENAME);
+    }
+
+    /**
+     * @return The path of the directory, where the AppMap Java agent should be stored (~/.appmap/lib/java).
+     */
+    @Nullable Path getOrCreateAgentDir() {
+        var agentDirPath = Paths.get(System.getProperty("user.dir")).resolve(Paths.get("appmap", "lib", "java"));
```
ahtrotta, line 164: This should be `.appmap` instead of `appmap`.

plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
```diff
@@ -0,0 +1,173 @@
...
+        // e.g. ~/.appmap/lib/java/appmap-1.7.2.jar
+        var assetDownloadPath = agentFilePath.resolveSibling(latestAsset.getFileName());
+
+        // don't download if the asset file does already exist
+        if (Files.exists(assetDownloadPath)) {
+            return false;
+        }
+
+        // e.g. appmap-1.7.2.downloading
+        var lockFilePath = assetDownloadPath.resolveSibling(latestAsset.getFileName().replace(".jar", ".downloading"));
```
ahtrotta, line 84: The lockfile is `appmap-X.X.X.jar.downloading`, not `appmap-X.X.X.downloading`, so you don't want to use the `replace` function. You can just add `.downloading` to the end of the file name.

plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
```diff
@@ -0,0 +1,173 @@
...
+        assert !assetDownloadPath.equals(lockFilePath);
+
+        if (!guardWithLockFile(lockFilePath, () -> downloadAgentJarFile(indicator, latestAsset, assetDownloadPath))) {
+            return false;
+        }
+
+        // create a symbolic link pointing to "appmap.jar"
+        try {
+            Files.deleteIfExists(agentFilePath);
+            Files.createSymbolicLink(agentFilePath, assetDownloadPath);
```
ahtrotta, line 94: This symbolic link points to an absolute file path:

image

Please make it so that it points to a relative path, like this:

image

ahtrotta (changes requested):

plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
jansorg: Thanks! This is fixed by  https://github.com/getappmap/appmap-intellij-plugin/pull/300/commits/0ead3e9c961bdf0a97243753352e42839d334b04

plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
jansorg: Fixed by https://github.com/getappmap/appmap-intellij-plugin/pull/300/commits/0ead3e9c961bdf0a97243753352e42839d334b04

plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
jansorg: Fixed by https://github.com/getappmap/appmap-intellij-plugin/pull/300/commits/0ead3e9c961bdf0a97243753352e42839d334b04

ahtrotta (approved):
Looks great, thanks!

PR #302, opened 2023-06-13
Disable update check of the IntelliJ Gradle plugin

jansorg:
Disable the build-time messages about available updates of the intellij-gradle plugin. This is purely to improve the development experience, because it's less cluttering the build output.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,5 +17,8 @@ ideVersion=IC-2021.3.3
 #ideVersion=IC-2022.3
 #ideVersion=IC-231.8109.90-EAP-SNAPSHOT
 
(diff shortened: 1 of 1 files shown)


ahtrotta:
I'm confused about this change. When and where do the unwanted messages occur? Will this affect our users at all? 

In the [link in the comment](https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin-build-features.html#selfupdatecheck) in the code, it says:

image

Why are we going against this advice?

jansorg:
The latest release is broken for me, e.g. because of https://youtrack.jetbrains.com/issue/IDEA-318263/Java-debugger-opens-.class-file-instead-of-the-projects-.java-source-file-when-a-breakpoint-is-hit.
Alternatively, we could just ignore the message until the Gradle plugin is fixed again...

PR #307, opened 2023-06-16
Don't show an empty list of findings after project is opened

jansorg:
The reload of findings was breaking because of a threading issue. 
The non-blocking ReadAction cancels the operation after the results are calculated. We used to call `onSuccess()`, which unfortunately may happen after or during the cancellation is triggered. This could cause an empty or an incomplete list of findings at startup.
This fix is hopefully fixing this for good. Unfortunately, I don't know of a reliable way to test this with a unit test.


kgilpin:
Maybe there is a way to use mocking to setup the race condition that you're
trying to test?

ahtrotta (approved):

Issue #299, opened 2023-06-12
Generated appmap.yml files should include appmap_dir


brikelly:
The extension should set the value of `appmap_dir` in the generated `appmap.yml` file when it is created. This will help our other tools to limit the scope / area of where they look.

PR #309, opened 2023-06-20
Set `appmap_dir` when creating an `appmap.yml` file

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/299

With this PR, property `appmap_dir` is added to a `appmap.yml` file before a run configuration is executed with "Execute with AppMap". Existing `appmap.yml` files, which don't have the property set, will be updated before the run configuration is executed.
If no path for the property was found, then it defaults to `tmp/appmap`.

Because the appmap.yml is always updated, the property is not passed to the Java agent.

Because there are now more possible exceptions before a run is executed, we're now displaying a notification instead of logging it as an error. Before, we were silently skipping some of the errors without telling the user about it.

- [x] Verify, that the `appmap_dir` path is below the working dir and does exist
- [x] Update existing `appmap.yml` file if it does not yet contain a `appmap_dir` property

### DIFF
--- a/plugin-core/src/main/java/appland/config/AppMapConfigFile.java
+++ b/plugin-core/src/main/java/appland/config/AppMapConfigFile.java
@@ -1,21 +1,112 @@
 package appland.config;
 
+import appland.utils.YamlUtils;
+import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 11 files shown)


jansorg:
Modify appmap.yml file

Issue #283, opened 2023-06-01
User can save filter settings and change the default filter


ahtrotta:

jansorg:
@ahtrotta Could you add a few details to help me understand what needs to be done? For example, I'm not sure which filter this is about.

brikelly:
@jansorg This was a new feature added to VS Code which is now released on that marketplace. There are some details here: https://github.com/getappmap/appmap-js/pull/1186

We are also creating a video to announce this feature so we can share that soon. It will demonstrate the behaviors and usage. Most of the work was done in appmap-js and already released, but the JetBrains plugin will need to add code to save filter settings appropriately.

ahtrotta:
@jansorg [here](https://docs.google.com/document/d/10uG5AtfYQst0Q-hq_hcT8yEzjk7qSZMeRC28bzeIiYA/edit) is a spec for this feature. Let me know if you have any questions!

jansorg:
@ahtrotta Would it be possible to keep `AppMapFilter` out of the frontend? As far as I understand, atm the frontend needs to know about the structure of `AppMapFilter` only to create the Base64 encoded state of the default filter. I can't import `AppMapFilter` as VSCode, but would have to copy (and keep in sync) the properties of this type (`rootObjects, `declutter` with all sub-properties).
IMHO the webview itself should init the default filter and just tell the host (i.e. the plugin) when it should be saved or updated. 
WDYT?

jansorg:
@ahtrotta Should `defaultFilter` only update the `default` flag or should it also update `state` and `filterName`? If it's the latter, what's the difference to `saveFilter`? Wouldn't `saveFilter` be enough to update the `defaultFilter` property?

ahtrotta:
@jansorg, I didn't think about the fact that you can't import `AppMapFilter` into JetBrains as we do in VS Code, so I'll make the frontend create and save a default filter if `savedFilters` is an empty array.

ahtrotta:
@jansorg `defaultFilter` should only update the `default` flag

ahtrotta:
[...]

PR #313, opened 2023-06-26
Support AppMap filters

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/283

Open issues:
- [x] On Linux, with IDEs < 2023.1.1, the keyboard input, e.g. to enter the name of a new filter, does not work. AFAIK there's no good fix available. Disabling off-screen rendering (which is causing the problem) has other side-effects, which are even worse. AFAIK this problem is not showing on Windows and macOS. As a possible solution, we could turn off the filters on Linux for < 2023.1.1. **Update**: Discussed with Brian to take action after the first feedback.


ahtrotta:
> * [ ]  On Linux, with IDEs < 2023.1.1, the keyboard input, e.g. to enter the name of a new filter, does not work. AFAIK there's no good fix available. Disabling off-screen rendering (which is causing the problem) has other side-effects, which are even worse. AFAIK this problem is not showing on Windows and macOS. As a possible solution, we could turn off the filters on Linux for < 2023.1.1

@brikelly @jansorg I just tested this and can confirm that no keyboard input is possible within the AppMap editor webview on IDE versions 2022.3.3 and 2023.1, but 2023.1.1 worked fine.

PR #322, opened 2023-07-06
Show AppMap icons for our webviews

jansorg:
This is a quick PR to show icons for our webview editors. Previously, ugly default icons were shown.
Now:
image


ahtrotta (approved):
There's a build error, but once that gets resolved it looks good!

jansorg:
I was glad for CI this time, because it caught a compatibility problem with a later version of the SDK. I've added a workaround.

ahtrotta (approved):

PR #331, opened 2023-07-21
Execute JVM patcher like the IDE does

jansorg:
Align our code with the platform, it does not introduce a functional change.


ahtrotta (commented):
@jansorg could you explain the context for this code change in more detail? I'm not sure exactly what the problem is/was and how the change addresses it.

jansorg:
Thanks for checking :) I found a flaw in this PR, i.e. we shouldn't make this change. I'll suggest a proper change in another PR, along with the description of the cause.

PR #334, opened 2023-07-26
Don't execute JDK validation on the EDT

jansorg:
During development and using `gradle runIde` I noticed frequent exceptions and also dead-locks of the UI when I executed the sprint-petclinic application with our "Execute with AppMap" button.
I don't know of a reliable case, but it happened to me mostly when I clicked on the button right after launching the IDE.

The exception does not contain our `appmap` package in the stacktrace, so user reports of it would have been sent to JetBrains and not AppMap.


ahtrotta (approved):
I noticed that the JDK wasn't properly configured often when using `./gradlew runIde`, so this is helpful, thanks!

Issue #298, opened 2023-06-12
"Install AppMap Agent" has a green checkmark in supported Java projects


brikelly:
Currently, the IntelliJ extension creates an `appmap.yml` in the project only when the user first starts their app or tests with AppMap (via the custom Run Configurations). 

We are planning to use the presence of `appmap.yml` in a project as a definitive sign that the user has successfully installed AppMap there. Therefore, we want the IntelliJ plugin to create the `appmap.yml` when they instead click this button to launch the "Install AppMap Agent" action:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/573108/93b804c1-f0f9-4a25-8937-83ac2608ba32)

jansorg:
- for Java only
- create for all Java modules if none was found

PR #338, opened 2023-07-27
Create appmap.yml after click on "Install AppMap Agent"

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/298

With this PR, a click on "Install AppMap Agent" creates a new appmap.yml file in a Java project.

Changes:
- Because the creation of `appmap.yml` was previously tied to run configurations, this PR has many changes to remove this limitation
- The top-level Java packages, which are collected for a new `appmap.yml` file, are now only restricted by a module and it's module dependencies. Previously, the search was restricted by the run configuration's search scope, which could have caused incomplete results. Because we can only locate the best possible module match for "Install AppMap agent", we can only restrict the package search with such module. Content of appmap.yml created from "Run with AppMap" and "Install AppMap Agent" should be equal.
- Disables "Run with AppMap" while indexing is in progress to avoid all kinds of bugs, e.g. empty Java packages in appmap.yml or dead-locks. Index access is needed to locate the Java packages for appmap.yml.
- Refactor the Java patching to run our own patching outside a ReadAction. Previously, with the SDK's Java patchers, we were using workarounds to run modifications inside the default ReadAction, but I noticed dead-locks caused by this.

I could move the last 2 items into a new PR if that's helpful for review.

Limitations:
- Because the click on "Install AppMap Agent" doesn't have file context, we can't create a new `appmap.yml` in all kinds of possible projects. Most notably, we're unable to create a appmap.yml file in a project with multiple independent modules, e.g. if it links to 2 more Gradle/Maven projects.


ahtrotta:
When I open a new project, then the instructions page opens immediately after the project loads and a new `appmap.yml` is created, which is the expected behavior:

image

But the `appmap.yml` is quite sparse, and has a blank string as the value for `appmap_dir`:

image

Is this expected? It would be good if we could generate the correct location for `appmap_dir`. 

Also, if I delete `appmap.yml`, then run some tests with AppMap, it will re-create `appmap.yml`, which is good:

image

The new `appmap.yml` has a list of `packages`, which were absent in the originally generated `appmap.yml` above. Could we generate these packages in the initial `appmap.yml` file?

jansorg:
The initial opening probably happened during indexing, which prevented the correct creation of appmap.yml. It‘s supposed to contain dir dir and packages.
Does deleting it and clicking „Install AppMp agent“ create a complete appmap.yml file for you?

ahtrotta:
Deleting it and then click `Install AppMap agent` does create a new `appmap.yml`, but it takes a very long time:

image


Also, the new `appmap.yml` still doesn't have a value for `appmap_dir`.

jansorg:
Hm, that sounds wrong. Is a progress bar displayed while it‘s created?

jansorg:
@ahtrotta Could you try again, please?
There was a bug, which used the wrong path as output directory, which would explain the empty value. I also added a fs refresh to speed up the display of the config file and added a background task progress to let you know what's going on.

I still need to solve the issue, that creating a config is not possible during indexing, e.g. when the Install Guide editor is opened right after the first project is opened.

jansorg:
@ahtrotta Thanks for testing! The reason for the use of indexed data is, that we need to collect the top-level Java packages (which are missing in your screenshot) and to determine the appmap dir, which is available only after a Gradle/Maven sync and indexing.
Would it be an option to not create appmap.yml when indexing is currently in progress? There may be a workaround for the build system detection, but there‘s certainly none to locate the Java packages. I‘d rather create correct data later or not all all instead of creating incorrect data, which will mess up the agent, indexer, etc.
We could also delay the creation of appmap.yml until indexing finished (e.g. indexing after the project was opened).

jansorg:
Update: It was decided to only display a green check mark icon instead of creating the underlying appmap.yml file when the instructions page is shown. If possible, a spinner icon should be displayed for item "Install AppMap Agent" in the sidebar.

jansorg:
Closing, the fundamental changes will be submitted a new PR.

Issue #329, opened 2023-07-21
"Sequence diagrams" should have a gray icon


jansorg:
![image](https://github.com/getappmap/appmap-intellij-plugin/assets/11473063/4381a964-28d6-41cb-80c1-a430755f28a1)

jansorg:
The different color is shown because the links has not been visited yet.

PR #342, opened 2023-07-31
Always show icons of install guide links in the same color

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/329

Now the slightly less colored icon is shown for install guide links.


jansorg:
CI seems to be unstable, but it‘s unrelated to this PR.

Issue #345, opened 2023-08-02
AppMaps fail to show up in the toolbar


ahtrotta:
On Linux, when AppMaps are created and indexed, they are NOT appearing in the AppMap toolbar:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/29b0c6b9-ccd5-403b-ac3f-9aa6c992ac13)

PR #347, opened 2023-08-04
Sync and watch AppMap directories with 2023.2

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/345, but this is only a partial fix

This adds AppMap output directories as watched roots and adds an indexable file set contributor, which is now the default extension point in 2023.2 to add (excluded) directories to the indexed locations.

This should fix missing items in the AppMap tool window with 2023.2.

### DIFF
--- a/plugin-core/src/main/java/appland/index/IndexUtil.java
+++ b/plugin-core/src/main/java/appland/index/IndexUtil.java
@@ -1,16 +1,89 @@
 package appland.index;
 
+import appland.config.AppMapConfigFile;
+import appland.files.AppMapFiles;
+import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.module.ModuleManager;
+import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.roots.ModuleRootManager;
+import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.concurrency.annotations.RequiresReadLock;
 import com.intellij.util.indexing.FileBasedIndex;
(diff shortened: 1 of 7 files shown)


plugin-core/src/main/java/appland/index/IndexUtil.java
```diff
@@ -1,17 +1,90 @@
...
+        }
+    }
+
+    /**
+     * Locate appmap.yml files without index access.
+     */
+    private static @NotNull Collection<VirtualFile> findConfigFilesByExcludedDirs(@NotNull Project project) {
+        var configFiles = VfsUtilCore.createCompactVirtualFileSet();
+        for (var module : ModuleManager.getInstance(project).getModules()) {
+            for (var root : ModuleRootManager.getInstance(module).getExcludeRoots()) {
```
dustinbyrne, line 63: Typically the `appmap.yml` will be located in the root of the project or a sub-project or sub-module. The path should typically not be git ignored. Maybe I'm misunderstanding, but what drives the need to locate an `appmap.yml` from the parent(s) of an excluded path?

jansorg: An excluded folder here is the exclusion by the IDE, which is not the same as .gitignore. Typically, build folders are excluded, e.g. „module-root/target“. This method is locating appmap.yml files without index access by using the excluded folders as starting point. But after thinking about this, this code is a bit stupid. I‘ll update this to look at content roots instead (i.e. module-root instead of „module-root/target“) as starting point.

jansorg: I pushed https://github.com/getappmap/appmap-intellij-plugin/pull/347/commits/04498a1be13638e9cf314b0786f2a27746c09a7a to fix this.

PR #351, opened 2023-08-10
Better fallbacks for AppMap output directory

jansorg:
Adds better fallback path for projects based on Maven or Gradle.

Especially Gradle projects have a complex structure. The root module does not have an output directory, only modules "main" or "test" have one attached (for spring-petclinic). Because we only receive the root module (without an output path), we were using "tmp/appmap" as fallback, but we need to use "target/appmap" for Gradle.

This adds a similar fallback for Maven based run configurations.

Unfortunately, I'm not aware of a way to run unit tests based on Gradle or Maven run configurations. The SDK's gradle plugin does something similar, but the test suite is not provided with the SDK and replicating it is out of scope (in my opinion).


dustinbyrne (approved):

dustinbyrne:
I've rebased this. Once the build passes it can be merged.

PR #363, opened 2023-08-21
Support `external-service` and `external-route`

### LINKED ISSUE #272: Class map item type "external-route" should be supported
See https://github.com/getappmap/appmap-intellij-plugin/issues/271 for a bug, where this unknown type triggered an exception

jansorg:
Closes #272


jansorg:
See https://github.com/getappmap/appmap-intellij-plugin/issues/272 for the note why this is closed.

PR #367, opened 2023-08-24
Initial GitHub Actions setup

jansorg:
Adds a basic workflow to run our tests on GitHub actions.
The branch is based on https://github.com/getappmap/appmap-intellij-plugin/pull/360, because `./gradlew test` should always be executed with the AppMap agent.

AppMaps are generated into `<project root>/build/appmaps`.

Because requests to api.github.com are rate-limited on GitHub Actions, I had to pass the GITHUB_TOKEN in the download requests, which downloads the agent.jar. This is only done on CI to avoid side-effects on users' machines.


build.gradle.kts
```diff
@@ -1,3 +1,6 @@
+import com.adarshr.gradle.testlogger.theme.ThemeType
```
kgilpin, line 1: Is it possible to use the documented Gradle support instead of this custom code? I'd like to understand the differences.

https://appmap.io/docs/reference/appmap-gradle-plugin.html#installation

.github/workflows/build.yml
```diff
@@ -0,0 +1,46 @@
...
+                    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                run: ./gradlew check
+
+            -   name: Verify plugin
+                shell: bash
+                run: ./gradlew verifyPlugin
+
+            -   name: Run Plugin Verifier
+                shell: bash
+                run: ./gradlew runPluginVerifier
```
kgilpin, line 37: Can you explain for me the difference between `verifyPlugin` and `runPluginVerifier`?

jansorg: `runPluginVerifier`: https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-runpluginverifier, it checks binary compatibility with the compatible major versions of the IDEs
`verifyPlugin`: https://plugins.jetbrains.com/docs/intellij/tools-gradle-intellij-plugin.html#tasks-verifyplugin, checks the deployment metadata

build.gradle.kts
jansorg: Let me try that. I didn't know that there's a AppMap Gradle  plugin

jansorg: I made a change for this at https://github.com/getappmap/appmap-intellij-plugin/pull/374, but had to drop the commit from this branch because the AppMap Gradle plugin is currently breaking our build and tests: https://github.com/getappmap/appmap-gradle-plugin/issues/18

PR #374, opened 2023-08-28
Use AppMap Gradle plugin to attach the agent

jansorg:
Based on https://github.com/getappmap/appmap-intellij-plugin/pull/367, which needs to be merged first

Tests are currently broken because of https://github.com/getappmap/appmap-gradle-plugin/issues/18


jansorg:
Closing, upstream has not been fixed yet and we have a working solution atm.

Issue #388, opened 2023-08-31
Findings should always be enabled if the user is signed in


jansorg:
The experimental setting should be removed

kgilpin:
@jansorg do you think that manual QA will be needed for this update?

jansorg:
@kgilpin That would be useful. Because something is removed, it's not possible to test well with instructions ;) But the general testing of the plugin onboarding should cover all of it (i.e. installing plugin for the first time, opening a new project, signing in to the AppMap server, record AppMaps).

PR #390, opened 2023-08-31
Always enable AppMap findings in the AppMap tool window

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/388

This PR removes the setting "Enable findings (experimental)" and always enables the support for AppMap findings:
- The AppMap scanner process is always launched, along with the AppMap indexer
- Telemetry messages `analysis:enable` and `analysis:disable` are not sent anymore, because the user can't toggle it anymore
- The Runtime Analysis panel does not display an empty text anymore. The empty text is shown if no node is visible in the tree. This doesn't happen anymore because of the "Findings overview" node. Previously, the node was hidden if the setting was turned off and a text with sign in instructions was shown.

### DIFF
--- a/plugin-core/src/main/java/appland/installGuide/InstallGuideEditor.java
+++ b/plugin-core/src/main/java/appland/installGuide/InstallGuideEditor.java
@@ -238,7 +238,6 @@ public class InstallGuideEditor extends WebviewEditor<List<ProjectMetadata>> {
         json.addProperty("page", type.getPageId());
         json.addProperty("userAuthenticated", settings.getApiKey() != null);
         json.addProperty("analysisEnabled", true);
-        json.addProperty("findingsEnabled", true);
     }
 
     private void handleMessageClickLink(@NotNull JsonObject message) {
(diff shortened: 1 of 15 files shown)


ahtrotta:
@jansorg could you rebase this when you get a chance? Thanks!

jansorg:
@ahtrotta It's rebased now.

plugin-core/src/main/java/appland/installGuide/InstallGuideEditor.java
```diff
@@ -242,8 +237,8 @@ private void addBaseProperties(@NotNull JsonObject json) {
 
         json.addProperty("page", type.getPageId());
         json.addProperty("userAuthenticated", settings.getApiKey() != null);
-        json.addProperty("analysisEnabled", settings.isAnalysisEnabled());
-        json.addProperty("findingsEnabled", settings.isEnableFindings());
+        json.addProperty("analysisEnabled", true);
```
ahtrotta, line 240: Rather than hard-coding this to `true`, can you just remove it?

jansorg: we could, but I wasn't sure what the webview would do with a missing property...

ahtrotta: Good point, I just checked and you can remove `findingsEnabled`, but don't remove `analysisEnabled`.

ahtrotta (approved):
The build is failing, but otherwise it looks good!

jansorg:
Merging now. A short instruction for testing was added at https://github.com/getappmap/appmap-intellij-plugin/issues/388

Issue #391, opened 2023-08-31
Onboarding instructions in JetBrains are missing information


ahtrotta:
Through testing, I've discovered that the problem arose from [this PR](https://github.com/getappmap/appmap-intellij-plugin/pull/376).

Using a fresh clone of `spring-petclinic`:

### VS Code

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/674a88c0-c4a5-4ba7-abcb-aeb30c07568e)

### JetBrains

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/2b7f8f59-5967-4148-b850-c34bc3d5205e)

### VS Code

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/2abfb618-c963-4d54-9112-6fef0311bdae)

### JetBrains

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/cb3e1e9f-ca48-4302-a6df-11e69a6cfcd9)

### VS Code

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/c6fb9cd9-718d-4401-b732-a07e91075578)

### JetBrains

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/7a04657f-ce69-4a25-a63e-4d1577d064a9)

PR #376, opened 2023-08-29
Execute webview callbacks in a background thread to avoid UI freezes on macOS

jansorg:
The handlers of messages posted from webview to IDE host are handled outside the EDT (the UI dispatch thread).
- On Linux, it's a regular background thread
- On macOS, it's `AWT-AppKit`, which is the thread handling native UI integration AFAIK.

On macOS, it happened that the EDT executed a WriteAction and a message handler of the Installation Guide webview on a ReadAction on thread `AWT-AppKit`. The handling on the EDT involved waiting on `AWT-AppKit`, which led to a deadlock.

This pull requests
1. refactors `InstallGuideEditor` to use our abstraction `WebviewEditor`. The other webviews were already using it and the install guide editor was the remaining implementation
2. refactors `WebviewEditor` to always execute webview message handlers in a background thread, which is not the calling thread of the handler, i.e. outside of `AWT-AppKit` on macOS

Thread dumps:
```
"AWT-EventQueue-0" prio=0 tid=0x0 nid=0x0 waiting on condition
     java.lang.Thread.State: WAITING
 on java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject@789a637e
	at java.base@11.0.13/jdk.internal.misc.Unsafe.park(Native Method)
	at java.base@11.0.13/java.util.concurrent.locks.LockSupport.park(LockSupport.java:194)
	at java.base@11.0.13/java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2081)
	at java.base@11.0.13/java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:433)
	at java.desktop@11.0.13/sun.awt.AWTThreading.execute(AWTThreading.java:147)
	at java.desktop@11.0.13/sun.awt.AWTThreading.executeWaitToolkit(AWTThreading.java:80)
	at java.desktop@11.0.13/sun.awt.AWTThreading.executeWaitToolkit(AWTThreading.java:53)
	at java.desktop@11.0.13/sun.awt.AWTThreading.executeWaitToolkit(AWTThreading.java:63)
	at java.desktop@11.0.13/sun.lwawt.macosx.CPlatformWindow.lambda$setVisible$10(CPlatformWindow.java:765)
	at java.desktop@11.0.13/sun.lwawt.macosx.CPlatformWindow$$Lambda$1187/0x0000000800cddc40.run(Unknown Source)
[...]

PR #393, opened 2023-08-31
Revert change which breaks status panel of install guide

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/391

This regression was caused by https://github.com/getappmap/appmap-intellij-plugin/pull/376. #376 refactored the webview handling to use shared code. To reduce differences between our 4 different webviews I synced appland-install-guide/webpack.config.js with appland/webpack.config.js, but that was wrong. It was wrong, because https://github.com/getappmap/appmap-intellij-plugin/commit/73f7801c47039f4a3b24c418fd2348e85cd50466#diff-cbccc897479584562e7f23f2aaa9137f949d80a1a615975ae42afa698a8cc9ba added the change to support dynamic properties, but apparently our other webviews don't need it.

This PR restores the support to properly render the status panel.


ahtrotta (approved):
This fixes it, thanks!

Issue #392, opened 2023-08-31
AppMap toolbar is blank


ahtrotta:
Through testing, I discovered that is bug arose from [this PR](https://github.com/getappmap/appmap-intellij-plugin/pull/376).

If I run `./gradlew clean runIde`, then the AppMap toolbar is blank. I navigated to `Tools --> AppMap --> Sign In To AppMap Server` to see if signing in would resolve the issue, but that was not possible because that option was disabled.

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/4ca8957e-a4cc-438a-810e-fd5e06596732)

PR #394, opened 2023-08-31
Revert change which broke the signin webview

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/392
https://github.com/getappmap/appmap-intellij-plugin/pull/376 removed the JS dist files of the signin webview, which is wrong.


ahtrotta (approved):
Sign-in is back, thanks!

Issue #397, opened 2023-09-04
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'value' of com/intellij/util/io/IOUtil.writeUTF must not be null


fabiangallardo:
Name | Value
----:|:-----
Plugin|appland.appmap
Version|0.48.0
IDE|IC-222.4167.29
OS|Windows 10 10.0
Last action|

### Stacktrace
```text
java.lang.IllegalArgumentException: Argument for @NotNull parameter 'value' of com/intellij/util/io/IOUtil.writeUTF must not be null
	at com.intellij.util.io.IOUtil.$$$reportNull$$$0(IOUtil.java)
	at com.intellij.util.io.IOUtil.writeUTF(IOUtil.java)
	at appland.index.BasicAppMapMetadataExternalizer.save(BasicAppMapMetadataExternalizer.java:31)
	at appland.index.BasicAppMapMetadataExternalizer.save(BasicAppMapMetadataExternalizer.java:12)
	at com.intellij.util.indexing.impl.ValueContainerImpl.saveTo(ValueContainerImpl.java:448)
	at com.intellij.util.indexing.impl.ChangeTrackingValueContainer.saveTo(ChangeTrackingValueContainer.java:173)
	at com.intellij.util.indexing.impl.ValueContainerExternalizer.save(ValueContainerExternalizer.java:23)
	at com.intellij.util.indexing.impl.ValueContainerExternalizer.save(ValueContainerExternalizer.java:12)
	at com.intellij.util.io.PersistentMapImpl.doPut(PersistentMapImpl.java:383)
	at com.intellij.util.io.PersistentMapImpl.put(PersistentMapImpl.java:365)
	at com.intellij.util.indexing.impl.ValueContainerMap.merge(ValueContainerMap.java:60)
	at com.intellij.util.indexing.impl.MapIndexStorage$1.onDropFromCache(MapIndexStorage.java:96)
	at com.intellij.util.indexing.impl.MapIndexStorage$1.onDropFromCache(MapIndexStorage.java:75)
	at com.intellij.util.containers.SLRUMap.clear(SLRUMap.java:166)
	at com.intellij.util.indexing.impl.storage.VfsAwareMapIndexStorage.processKeys(VfsAwareMapIndexStorage.java:113)
	at com.intellij.util.indexing.impl.storage.TransientChangesIndexStorage.processKeys(TransientChangesIndexStorage.java:139)
[...]

PR #399, opened 2023-09-06
Fix NPE caused by a `null` value for the name property of AppMap metadata.json

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/397

This fixes a user-reported exception. I'm not entirely sure how `name` of metadata.json could become empty. This PR supports the `null` value during indexing to now throw an exception.

This PR adds a test and a few necessary modifications to allow creation of AppMap metadata.json with a `null` name property value in our tests.

### DIFF
--- a/plugin-core/src/main/java/appland/index/BasicAppMapMetadataExternalizer.java
+++ b/plugin-core/src/main/java/appland/index/BasicAppMapMetadataExternalizer.java
@@ -10,13 +10,17 @@ import java.io.DataOutput;
 import java.io.IOException;
 
 class BasicAppMapMetadataExternalizer implements DataExternalizer<BasicAppMapMetadata> {
-    public static final BasicAppMapMetadataExternalizer INSTANCE = new BasicAppMapMetadataExternalizer();
+    static final @NotNull BasicAppMapMetadataExternalizer INSTANCE = new BasicAppMapMetadataExternalizer();
+
+    private BasicAppMapMetadataExternalizer() {
+    }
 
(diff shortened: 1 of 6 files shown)


plugin-core/src/main/java/appland/index/BasicAppMapMetadataExternalizer.java
```diff
@@ -10,13 +10,17 @@
...
+    }
 
     @Override
     public BasicAppMapMetadata read(@NotNull DataInput in) throws IOException {
-        var name = IOUtil.readUTF(in);
-        var hasTestStatus = in.readBoolean();
+        var hasName = in.readBoolean();
+        var name = hasName ? IOUtil.readUTF(in) : null;
 
+        var hasTestStatus = in.readBoolean();
```
ahtrotta, line 23: Correct me if I'm wrong, but aren't `hasName` and `hasTestStatus` always going to be equal to each other because they are both the return value from `in.readBoolean()`? That would mean that `name` and `jsonId` would always be equal to each other, too. Is that intentional?

jansorg: ˋinˋ represents a stream of data (bytes), and ˋreadBooleanˋ has side-effects, because it reads from the stream and advances the current position. Calling it twice is reading to different values.

ahtrotta (approved):
It looks like the Travis build is failing, but otherwise looks good, thanks!

Issue #385, opened 2023-08-31
appmap.yml must not be overridden with new config setup


jansorg:
appmap.yml seems to be overwritten when "Run with AppMap" is launched.

brikelly:
@jansorg this still seems to be a problem. If I make changes to `appmap.yml`, they disappear after I choose "Start with AppMap".

jansorg:
@brikelly Could you clarify which changes you made to the appmap.yml file, which were removed?

brikelly:
@jansorg here's a Loom showing the issue: https://www.loom.com/share/506a906965f641e4b13dd1090b67daf2?sid=d5cbc802-bc9f-457c-8f29-01a2c9da205c

PR #409, opened 2023-09-08
Don't override appmap.yml with new values when it's updated

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/385

`appmap.yml` was overridden with a new configuration because the search scope to locate existing configurations did not contain all content roots of a module. 

This PR fixes the bug and adds test for relative and absolute input paths to define the appmap_dir value. The `appmap_dir` value is always output as a relative path, as before. This PR also corrects an issue with the handling of relative path values, which was discovered by the new tests.


apotterri (approved):
LGTM

brikelly:
@jansorg I still see the file getting overridden in testing. Are there circumstances where you still expect this to happen?

Issue #366, opened 2023-08-22
Use the new stdout messages from the indexer to simplify file watching


brikelly:
The changes to the indexer were published recently and now it streams status to stdout. The IntelliJ plugin should use those now for file modification watches.

jansorg:
Depends on https://github.com/getappmap/appmap-js/pull/1317

PR #417, opened 2023-09-12
Refresh local filesystem after AppMap indexes were create or updated

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/366

This PR contains tests for the parsing of `Indexed` events.

The plugin now listens to `Indexed ` events emitted by the AppMap indexer tool. 
This is used to detect new or changed AppMap indexes more reliably. The existing implementation to convince the IDE to index AppMap data is left untouched.

The process listener triggers an async refresh of the file cache and data based on it, e.g. of our file-based indexes.

### DIFF
--- a/plugin-core/src/main/java/appland/cli/IndexerEventUtil.java
+++ b/plugin-core/src/main/java/appland/cli/IndexerEventUtil.java
@@ -1,0 +1,36 @@
+package appland.cli;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Parses the "Indexed" messages documented at
+ * <a href="https://github.com/getappmap/appmap-js/blob/main/packages/cli/doc/index-verbose.md">index-verbose.md</a>.
+ */
+class IndexerEventUtil {
+    private IndexerEventUtil() {
+    }
+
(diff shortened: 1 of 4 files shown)


plugin-core/src/main/java/appland/cli/IndexerEventUtil.java
```diff
@@ -0,0 +1,36 @@
...
+                ? unescapeFilePath(value.substring(1, value.length() - 1))
+                : value;
+    }
+
+    private static @NotNull String unescapeFilePath(@NotNull String value) {
+        return value
+                .replace("\\\\", "\\")
+                .replace("\\n", "\n")
+                .replace("\\0", "\0")
+                .replace("\\\"", "\"");
```
dividedmind, line 34: This is wrong, it would unescape `\\n` into a newline. Replacing double-backslash should come last.

jansorg: Good point. The problem is that multiple replacements in the whole string creates ambiguities. Even with replacing double-backslash last something like "backslash-backslash-n" `\\\\n` would be unescaped into `backslash-newline`. I'll fix it with a one-time replacement using a regex.

Issue #424, opened 2023-09-14
Telemetry for findings should be removed


brikelly:
Specifically, this blob should be removed from the plugin: https://github.com/getappmap/appmap-intellij-plugin/blob/17fa0bd14f1d5e7a218bad461c56d7b2ce5baadc/plugin-core/src/main/java/appland/webviews/findings/FindingsOverviewEditor.java#L55-L87

kgilpin:
You're just going to remove that one file? Or you're going to do a telemetry cleanup like in VSCode?

There is already a story for that BTW.

Issue #398, opened 2023-09-05
Refactor telemetry to match VSCode


kgilpin:
https://github.com/getappmap/vscode-appland/pull/790 refactored VSCode telemetry to remove unreliable events and focus on error reporting. We should roll these changes over to JetBrains as well.

jansorg:
Based on https://github.com/getappmap/appmap-intellij-plugin/issues/424, the following events would be removed:
- view:open
- click_docs_link_in_tree
- clicked_sign_in_link
- click_tab
- select_object_in_sidebar
- analysis:view_finding
- analysis:view_overview
- analysis:cta_interaction
- authentication:success
- authentication:sign_out
- plugin:install
- appmap:create
- appmap:open
- sidebar_search_focused
- click_filter_button
- reset_diagram

The following would remain:
- run_config:incompatible_jdk (only sent by the JetBrains plugin, I suppose)
- run_config:execute (only sent by the JetBrains plugin, I suppose)
- authentication:failed
- open_api:failure

`project:open` is currently NOT sent.

jansorg:
In general, exceptions by the plugin are sent as GitHub issues after the user confirmed to send it. That's the way the JetBrains IDE's handle this (i.e. the manual confirmation before sending error reports to a server of the plugin's choice).


Possible errors:
- `authentication:failed`: Failure during/after sign in. This may indicate a problem with our sign in code, but there's no further context. It could also happen for server errors, I assume.
- `open_api:failure`: Sent if the AppMap CLI to generate OpenAPI did not return with exit code 0. It may indicate a problem with our code, e.g. when the CLI tool changed the supported command line.

No errors in our code:
- `run_config:incompatible_jdk`: This is about a problem with user's project. So this isn't a problem in our code.
- `run_config:execute`: A status message about run configuration execution. I wasn't sure if it should be removed, because it wasn't part of the VSCode refactor.

kgilpin:
Ok thanks. Let’s drop `run_config:execute` and keep the other ones because they might be bugs in our code.

PR #430, opened 2023-09-15
Remove telemetry except of messages, which may indicate a bug in our code

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/424
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/398

This PR removes most of the telemetry message. Please refer to [this comment](https://github.com/getappmap/appmap-intellij-plugin/issues/398#issuecomment-1721241742) for a list of the removed messages.

### DIFF
--- a/plugin-core/src/main/java/appland/webviews/findings/FindingsOverviewEditor.java
+++ b/plugin-core/src/main/java/appland/webviews/findings/FindingsOverviewEditor.java
@@ -6,7 +6,6 @@ import appland.problemsView.FindingsManager;
 import appland.problemsView.FindingsViewTab;
 import appland.problemsView.model.ImpactDomain;
 import appland.problemsView.model.ScannerFinding;
-import appland.telemetry.TelemetryService;
 import appland.utils.GsonUtils;
 import appland.webviews.WebviewEditor;
 import appland.webviews.findingDetails.FindingDetailsEditorProvider;
@@ -15,7 +14,6 @@ import com.google.gson.JsonObject;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.Messages;
-import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import it.unimi.dsi.fastutil.objects.Object2IntMap;
 import it.unimi.dsi.fastutil.objects.Object2IntMaps;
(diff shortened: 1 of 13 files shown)


plugin-gradle/src/main/java/appland/execution/AppMapGradleAgentRunner.java
```diff
@@ -34,11 +34,7 @@ public boolean canRun(@NotNull String executorId, @NotNull RunProfile profile) {
     @Override
     public void execute(@NotNull ExecutionEnvironment environment) throws ExecutionException {
         verifyJdk(environment);
-        try {
-            findDelegate().execute(environment);
-        } finally {
-            AppMapJvmExecutor.sendTelemetry(environment.getRunProfile());
-        }
+        findDelegate().execute(environment);
```
ahtrotta, line 37: Why is this no longer in a `try` block? If it raises an exception, is it a problem that we are no longer catching it?

plugin-java/src/main/java/appland/execution/AbstractAppMapJavaAgentRunner.java
```diff
@@ -27,11 +27,7 @@ public abstract class AbstractAppMapJavaAgentRunner extends DefaultJavaProgramRu
     @Override
     public void execute(@NotNull ExecutionEnvironment environment) throws ExecutionException {
         verifyJdk(environment);
-        try {
-            super.execute(environment);
-        } finally {
-            AppMapJvmExecutor.sendTelemetry(environment.getRunProfile());
-        }
+        super.execute(environment);
```
ahtrotta, line 30: As above, why is this no longer in a `try` block?

plugin-core/src/main/java/appland/webviews/findings/FindingsOverviewEditor.java
```diff
@@ -53,37 +51,6 @@ protected void setupInitMessage(@Nullable List<ScannerFinding> initData, @NotNul
...
-
-            uniqueImpactDomainCounts.forEach((domain, count) -> {
-                if (count > 0) {
-                    event.metric("num_findings.num_unique_" + domain.getJsonId().toLowerCase(), count.doubleValue());
-                }
-            });
-
-            return event;
-        });
     }
```
ahtrotta, line 54: It looks like this is now an empty method, should we just remove the method entirely?

ahtrotta (commented):
Are there automated tests that check if we are sending error telemetry? If not, could you add tests for that?

plugin-gradle/src/main/java/appland/execution/AppMapGradleAgentRunner.java
jansorg: Try-finally means that exceptions are passed to the caller (i.e. they are not caught), but the finally code is always executed, even when an exception was thrown by the try block. 
we don‘t have code in finally anymore, so it was removed.

ahtrotta: Okay, makes sense, thank you!

jansorg:
There are no tests for the remaining messages:
- `run_config:incompatible_jdk`
- `authentication:failed`
- `open_api:failure`

An incompatible JDK setup can't be tested well. Testing the remaining two messages would require a mock OAuth server setup and a mocked AppMap CLI, which would take considerable time. Could you please open a new issue if you'd like tests?

ahtrotta (approved):
I think if the tests are going to require that much effort then we can hold off on them for now. This looks good, thanks!

Issue #432, opened 2023-09-16
Users see a notification box when they create their first map


brikelly:
Similar to these types of message boxes:
![Image](https://user-images.githubusercontent.com/573108/268423788-1c1bda4a-39bf-4f79-adb3-568925adc193.png)

We need to show a similar type of notification message when a user creates their very first map after installing the plugin for the first time, and only show it that one time. The purpose is to help guide new users towards their AppMaps, because they might not notice that they've been created or know how to even open them.

The text of the notification should be:
"You've created your first AppMap! Congratulations."
There should be one button as a primary action, colored in blue as above, with the text "Open your new AppMap"

This notification should only be seen at most once on a user's system.

jansorg:
> This notification should only be seen at most once on a user's system.
@brikelly This requirement doesn't fit with the way settings are handled. What's possible is to show only once for the installation of an IDE. For example, a user using IntelliJ Ultimate would see it only once on a system for IntelliJ Ultimate. But if the same user has installed another IDE, e.g. RubyMine, then the notification would be displayed once for RubyMine, too.
Would this be good enough?

brikelly:
> > This notification should only be seen at most once on a user's system.
> 
> 
> 
> @brikelly This requirement doesn't fit with the way settings are handled. What's possible is to show only once for the installation of an IDE. For example, a user using IntelliJ Ultimate would see it only once on a system for IntelliJ Ultimate. But if the same user has installed another IDE, e.g. RubyMine, then the notification would be displayed once for RubyMine, too.
> 
> Would this be good enough?

Yes, absolutely fine to do it this way and show it once per installation.

PR #435, opened 2023-09-18
Notifiy user about first AppMap

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/432

Displays a notification when the first AppMap was created.

- It does not check if the user is signed in
- The notification is NOT shown for existing users (`isFirstStart == false`) 
- It displays the "Explore AppMaps" page when the user clicks on the link in the notification
- The notification is only displayed when a new AppMap file is created while the matching project is opened with the AppMap plugin. If a project already contains AppMap files at first startup, then no notification is displayed for these AppMaps.

Possible changes:
- We could display a title and AppMap icon for the notification. Currently, only the message content is shown.

### DIFF
--- a/plugin-core/src/main/java/appland/files/AppMapFileChangeListener.java
+++ b/plugin-core/src/main/java/appland/files/AppMapFileChangeListener.java
@@ -2,6 +2,7 @@ package appland.files;
 
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.util.messages.Topic;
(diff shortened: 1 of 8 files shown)


ahtrotta (approved):
Code looks good and I tested this manually with a fresh clone of petclinic and it worked well, thanks!

I'm going to approve and merge this because we're under time pressure, but if it's feasible to add automated tests, can you add a new issue for adding automated tests for this feature?

PR #437, opened 2023-09-18
Refresh parent directory of new AppMaps for better user experience

### LINKED ISSUE #436: Refresh after CLI notification must also refresh for the new metadata directory
Currently, only the path of the `.appmap.json` file is refreshed. But the metadata directory of the AppMap must be refreshed, too.

jansorg:
Closes #436 
Refreshing the parent directory triggers earlier (re)indexing of the AppMap metadata directories.
This results in a quicker UI update of available AppMaps, because the list of AppMaps relies on the metadata indexes.


ahtrotta (commented):
Would it be possible to add a test for this?

jansorg:
I added a test.

ahtrotta (approved):
Great, thank you for adding the test! This looks good, it looks like you'll need to rebase to `develop` before merging, though.

PR #445, opened 2023-09-19
Properly unescape indexer file paths

jansorg:
Resolves the issue reported at https://github.com/getappmap/appmap-intellij-plugin/pull/417#pullrequestreview-1632257758

Multiple rounds of replacements create ambiguities, we're now iterating the string only once and remember if the last char was escaped or not.

### DIFF
--- a/plugin-core/src/main/java/appland/cli/IndexerEventUtil.java
+++ b/plugin-core/src/main/java/appland/cli/IndexerEventUtil.java
@@ -27,10 +27,36 @@ class IndexerEventUtil {
     }
 
     private static @NotNull String unescapeFilePath(@NotNull String value) {
-        return value
-                .replace("\\\\", "\\")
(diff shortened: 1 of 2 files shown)


dividedmind (approved):
This looks ok, thanks!

plugin-core/src/main/java/appland/cli/IndexerEventUtil.java
```diff
@@ -27,10 +27,36 @@ static boolean isIndexedEvent(@NotNull String message) {
...
+                        break;
+                    case 'n':
+                        result.append('\n');
+                        break;
+                    case '0':
+                        result.append('\0');
+                        break;
+                    case '"':
+                        result.append('"');
+                        break;
```
dividedmind, line 54: You could remove the `\` and `"` cases by adding `default: result.append(c)`.

PR #448, opened 2023-09-19
Open installation instructions after first sign in

### LINKED ISSUE #446: Show "Install AppMap Agent" page after sign-in
- When they first install the plugin, the side panel needs to be expanded so that they see the Sign In button clearly (this is working :+1: )
- When they sign in via their browser and come back to IntelliJ, the "Install AppMap Agent" page should be opened (not working)

cc @brikelly

jansorg:
Closes #446


ahtrotta (approved):
This looks good to me and it worked well when I tested it manually, but it looks like we'll need to rebase this branch with `develop` before we merge

Issue #421, opened 2023-09-13
AppMaps are organized by category (recording method) similar to how VS Code does


brikelly:
Needs detail - should be same as VS Code

kgilpin:
He also wants the timestamps to be:

a) Consistent format eg ISO 8601 both in filenames and in the UI
b) Any date/time display in `ago` format for the last 24 hours - e.g. `1 minute ago`, `22 minutes ago`, `1 hour ago` etc

brikelly:
@jansorg can you make this your next priority for this week?

kgilpin:
Example:

<img width="428" alt="Screen Shot 2023-09-18 at 2 12 51 PM" src="https://github.com/getappmap/appmap-intellij-plugin/assets/86395/112d8ca2-a60f-4b8e-8a1a-019a239a406a">

jansorg:
- Where should the timestamps be displayed? VSCode is showing the time only (including milliseconds!) if it's the current day (I think). Otherwise it's showing the date.
- Are timestamps only displayed for requests or also for items like queries?

![image](https://github.com/getappmap/appmap-intellij-plugin/assets/11473063/fe0aab82-1b20-4a1e-9d28-9d14df8809a1)

kgilpin:
> How is VSCode building the tree hierarchy for AppMaps?

By project name, recorder type, recorder name. This information is available in the [AppMap metadata](https://github.com/getappmap/appmap#metadata) as recorder type and recorder name.

> Where should the timestamps be displayed? VSCode is showing the time only (including milliseconds!) if it's the current day (I think). Otherwise it's showing the date.

It's OK with me to follow the VSCode format since no-one has complained about it - yet :-)

> Are timestamps only displayed for requests or also for items like queries?

Timestamps can be shown for request recordings and remote recordings.

jansorg:
@kgilpin How is "java"  in the screenshot above retrieved? Is it "language.name" of the metadata? If it's null, what's displayed instead?

jansorg:
Is the timestamp the "last modified" time of the AppMap .appmap.json file?

PR #452, opened 2023-09-20
Display AppMap hierarchy

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/421

AppMaps are now displayed in a hierarchy:
- `<module name>`
  - `<recorder type> (<language name> + <recorder name>`)
    - `<AppMap name>`
- The tree hierarchy is expanded by default 
- The indexed data of AppMaps had to be enhanced to provide the data needed to group AppMaps
- Adds unit tests for empty AppMaps, AppMaps in one module and AppMaps in two modules of a project

image

### DIFF
--- a/plugin-core/src/main/java/appland/index/AppMapMetadata.java
+++ b/plugin-core/src/main/java/appland/index/AppMapMetadata.java
@@ -21,6 +21,9 @@ public final class AppMapMetadata {
      */
     private final @NotNull String name;
     private final @Nullable TestStatus testStatus;
+    private final @Nullable String languageName;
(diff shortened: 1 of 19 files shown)


ahtrotta:
I've been trying to get this to work locally by using the `runIde` command, but the extension isn't running the `index --watch` and `scan --watch` commands:

image

ahtrotta:
@jansorg this seems to be a problem on `develop`, so I think it might be an issue with my machine. I'll try to figure it out.

ahtrotta (approved):

PR #455, opened 2023-09-27
Don't override properties of an existing appmap.yml file

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/385

The appmap.yml wasn't properly located if the executed run configuration defined its own search scope.
The test was updated to make the previous implementation fail as  expected.
Now, the search scope is just the content search scope of the module.


ahtrotta (approved):
Thanks Joachim!

Issue #458, opened 2023-09-27
Clicking on a code object (SQL/API) in the code objects view does not select the location in the Dependency Map where that object is used.


petecheslock:
Clicking on a code object (SQL/API) in the code objects view does not select the location in the Dependency Map where that object is used.

[Here is a video showing the issue](https://www.loom.com/share/53bcb74832224bb0b4c455b9583e7911)

This map is the one used in the demo video.  
[New_Group_Creation.appmap.json.zip](https://github.com/getappmap/appmap-intellij-plugin/files/12742589/New_Group_Creation.appmap.json.zip)

When clicking on a SQL query or an API request, the code editor should open up the relevant appmap and select the location on the dependency map where that code object is used.  Currently in Intellij -the appmap is opened but the code object is not selected in the viewer.  

See this example from VS Code showing the expected outcome.  

<img width="1649" alt="Screenshot 2023-09-27 at 3 26 12 PM" src="https://github.com/getappmap/appmap-intellij-plugin/assets/511733/54d71d49-37d5-4bfc-a75e-488a4b68385b">

brikelly:
@jansorg This is actually more important than the other IntelliJ tasks we discussed today - can we prioritize this one at the top of your to-do list please? Thanks!

brikelly:
And if you need the source code to the app, this is it I believe: https://github.com/finos/waltz

PR #460, opened 2023-09-28
Open query and request code objects in AppMap view

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/458

The fqid of query and request code objects was calculated incorrectly.

There were tests, but the test fixture was taken from the VSCode plugin at https://github.com/getappmap/vscode-appland/blob/5051cd95fc8732e96f11021ce7ed6acba714b9c4/test/integration/buildClassMap.test.ts#L9.
This test data did not contain fqids of queries or requests and our own tests did not detect incorrect fqids.

This PR fixes the tests data and the implementation to provide the correct fqids for query and request code objects.
The fqids are passed to the AppMap view to navigate to such item.


ahtrotta (approved):
Looks good and works well, thanks!

Issue #407, opened 2023-09-07
Opening an AppMap from a Findings webview page will open the AppMap to the Finding


ahtrotta:
### Issue

Currently, when opening an AppMap from a Findings webview page:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/668f736e-8ac9-4c09-8da5-e3c755044e98)

The AppMap will open to the Trace View:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/4e9a3059-c141-4c64-bcde-f933d1f763bd)

What we want is for the AppMap to open to the Finding:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/b89cb45b-564e-4f64-a4a6-151b19c0dd0b)

### Solution

To do this, change the `fragment` property [when opening the AppMap](https://github.com/getappmap/appmap-intellij-plugin/blob/develop/plugin-core/src/main/java/appland/webviews/findingDetails/FindingDetailsEditor.java#L69) to be `{ selectedObject: analysis-finding:{{ HASH_OF_FINDING }} }`. For the change in VS Code, see [here](https://github.com/getappmap/vscode-appland/commit/7c0a156e03b8b25bcf8acd29df7c4cf06db29d38).

PR #471, opened 2023-10-11
Center AppMap webview on finding

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/407

- Keep the source finding hash in the finding details editor
- Use this hash to create the editor state for the new AppMap webview
- Use the new state to center the newly opened AppMap editor on the source finding

(@ahtrotta Why wasn't the fragment changed, which is passed from finding webview to the webview host? I think this would have changed the behavior in both VSCode and JetBrains without changes to the plugins)

## Testing
This PR  needs manual testing to verify the navigation

### DIFF
--- a/plugin-core/src/main/java/appland/problemsView/FindingsManager.java
+++ b/plugin-core/src/main/java/appland/problemsView/FindingsManager.java
@@ -297,14 +297,24 @@ public class FindingsManager implements ProblemsProvider {
 
     private void loadFileLocked(@NotNull VirtualFile findingsFile, @NotNull Consumer<Problem> notifier) {
         var fileData = loadFindingsFile(findingsFile);
-        if (fileData != null && fileData.findings != null) {
-            for (var finding : fileData.findings) {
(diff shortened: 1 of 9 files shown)


ahtrotta:
I confirmed that the `N plus 1 SQL query` findings in `spring-petclinic` do not open with the correct events highlighted in the flamegraph. However, this feature works correctly in the VS Code extension with the same project. Can you look into why that might be?

jansorg:
@ahtrotta I've pushed https://github.com/getappmap/appmap-intellij-plugin/pull/471/commits/cc7ac9af3b34f699daa33f52598b89f926d70026 to fix the difference in rendering. We're now passing the original JSON data of the finding to the AppMap webview and not just a subset.

ahtrotta (approved):
Looking good!

jansorg:
Merging, the CI error is caused by the AppMap agent and unrelated to the changes of this PR.

PR #472, opened 2023-10-11
Sort AppMaps of request or remote recording by modification date

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/459

AppMaps, which are located in remote_recording or request_recording groups, are ordered by modification date.
The newest AppMaps appear first in the list.

This PR adds a test to verify the sort order by modification date.

- [x] This PR is using the `recorder.name` property of the AppMap metadata to decide if it needs sorting. Is this correct or should a property like `recorder.type`  be used instead? Which would be the names to check for?


ahtrotta:
With regards to `recorder.name` vs `recorder.type`, please change how this feature works to use `recorder.type` since it should be more stable and only has limited options for its value (see the [spec](https://github.com/getappmap/appmap#example)).

ahtrotta (approved):
Thanks @jansorg!

Issue #486, opened 2023-10-26
com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: Read access is allowed from inside read-action or Event Dispatch Thread (EDT) only (see Application.runReadAction()); see https://jb.gg/ij-platform-threading for details


aakil786:
Name | Value
----:|:-----
Plugin|appland.appmap
Version|0.51.0
IDE|IU-232.10072.27
OS|Linux 6.2.0-35-generic
Last action|

### Stacktrace
```text
com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: Read access is allowed from inside read-action or Event Dispatch Thread (EDT) only (see Application.runReadAction()); see https://jb.gg/ij-platform-threading for details
Current thread: Thread[ApplicationImpl pooled thread 19,4,main] 401318903 (EventQueue.isDispatchThread()=false)
SystemEventQueueThread: Thread[AWT-EventQueue-0,6,main] 1066394419
	at com.intellij.openapi.application.impl.ApplicationImpl.createThreadAccessException(ApplicationImpl.java:1083)
	at com.intellij.openapi.application.impl.ApplicationImpl.assertReadAccessAllowed(ApplicationImpl.java:1038)
	at com.intellij.psi.impl.file.impl.FileManagerImpl.findFile(FileManagerImpl.java:351)
	at com.intellij.psi.impl.PsiManagerImpl.findFile(PsiManagerImpl.java:156)
	at com.intellij.ide.impl.dataRules.PsiFileRule.getData(PsiFileRule.java:23)
	at com.intellij.ide.impl.DataManagerImpl.getRulesData(DataManagerImpl.java:210)
	at com.intellij.ide.impl.DataManagerImpl.lambda$getDataRuleInner$8(DataManagerImpl.java:183)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromRuleInner(DataManagerImpl.java:112)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromRulesInner(DataManagerImpl.java:99)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromProviderAndRulesInner(DataManagerImpl.java:85)
	at com.intellij.ide.impl.DataManagerImpl.lambda$getDataFromRuleInner$3(DataManagerImpl.java:113)
	at com.intellij.openapi.actionSystem.DataKey.getData(DataKey.java:68)
[...]

PR #491, opened 2023-11-06
Fetch Java run configuration parameters under ReadAction

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/486

There's existing test coverage to test that the `verifyJdk` method doesn't fail.


jansorg:
This isn't the right fix, refer to https://youtrack.jetbrains.com/issue/IDEA-337277/SpringBootCommandLineState-expands-macros-without-required-ReadAction to details.

Issue #490, opened 2023-11-05
com.intellij.diagnostic.PluginException: PSI element is provided on EDT by appland.toolwindow.appmap.AppMapWindowPanel.getData("psi.File"). Please move that to a BGT data provider using PlatformCoreDataKeys.BGT_DATA_PROVIDER [Plugin: appland.appmap]


beingrohit-exe:
Name | Value
----:|:-----
Plugin|appland.appmap
Version|0.52.0
IDE|IU-232.10203.10
OS|Windows 11.0
Last action|

### Stacktrace
```text
com.intellij.diagnostic.PluginException: PSI element is provided on EDT by appland.toolwindow.appmap.AppMapWindowPanel.getData("psi.File"). Please move that to a BGT data provider using PlatformCoreDataKeys.BGT_DATA_PROVIDER [Plugin: appland.appmap]
	at com.intellij.diagnostic.PluginProblemReporterImpl.createPluginExceptionByClass(PluginProblemReporterImpl.java:23)
	at com.intellij.diagnostic.PluginException.createByClass(PluginException.java:89)
	at com.intellij.ide.impl.DataValidators.reportPsiElementOnEdt(DataValidators.java:96)
	at com.intellij.ide.impl.DataValidators.isDataValid(DataValidators.java:71)
	at com.intellij.ide.impl.DataValidators.validOrNull(DataValidators.java:64)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromProviderInner(DataManagerImpl.java:244)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromProviderAndRulesInner(DataManagerImpl.java:81)
	at com.intellij.ide.impl.DataManagerImpl.getDataFromProviderAndRules(DataManagerImpl.java:60)
	at com.intellij.openapi.actionSystem.impl.PreCachedDataContext.cacheProviderData(PreCachedDataContext.java:332)
	at com.intellij.openapi.actionSystem.impl.PreCachedDataContext.cacheComponentsData(PreCachedDataContext.java:308)
	at com.intellij.openapi.actionSystem.impl.PreCachedDataContext.<init>(PreCachedDataContext.java:101)
	at com.intellij.openapi.actionSystem.impl.Utils.newPreCachedDataContext(Utils.java:122)
	at com.intellij.openapi.actionSystem.impl.Utils.wrapToAsyncDataContext(Utils.java:104)
[...]

PR #493, opened 2023-11-06
Enforce that PsiFile is fetched with a slow/background data provider

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/490

Fixes exceptions on newer SDKs. It seems like the slow data provider is only queried if the fast provider does not provide a result.
This PR is always returning `null` when `PsiFile` is requested to enforce this.

I've tested this manually with 2021.3 and 2023.2.


ahtrotta (approved):
I was able to reproduce the error on `develop` and I was able to verify that this fixes it. Thanks!

Issue #411, opened 2023-09-11
AppMap CLI process could be terminated in a better way


jansorg:
The service, which keeps track of running AppMap CLI processes, is terminating them when the project is closed.
It currently removes the process entry from the internal map and then terminates the 2 linked AppMap processes. 
This doesn't support the following two cases well:
- AppMap processes crash (e.g. because of a bug). In this case the process is not detected as terminated and no restart is attempted (if we'd like to do this)
- Processes, which don't terminate as expected, can't be detected.

kgilpin:
There is a restart with back off in VSCode extension

PR #497, opened 2023-11-15
Restart AppMap process when it terminates unexpectedly

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/411

This PR implements restart of AppMap CLI processes.
Initially, when processes are launched for an AppMap-enabled directory, both indexer and scanner processes are started.
Now, when one of the processes terminates unexpectedly, this process is restarted after a delay. The initial delay is `5s`, the next next `5s*1.5=7.25s` and the next `7.25s*1.5=10.9s` and after the third unexpected termination no further restart is attempted.

This PR adds a tests to verify the restart.


ahtrotta (approved):
Worked well when I gave it a spin and looks good, thanks!

Issue #492, opened 2023-11-06
Attempt to remove usage of internal API


jansorg:
JetBrains is restricting the use of internal API and may reject plugin updates in the future if it's still using such internal API.
We're using internal API in a few places and should attempt to replace them with support API calls.

Usage of internal API is shown on the Marketplace page in the section of the plugin verifier results:
https://plugins.jetbrains.com/plugin/16701-appmap/edit/versions/stable/429382

PR #498, opened 2023-11-15
Remove usage of internal API

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/492

Replaces usage of all `@Internal` API with an alternative. We're not using internal API anymore.
This PR also drops use of API which was deprecated in 2021.3, which is the earliest supported version.

The changes to drop internal API from AppMapFilesIterator have the side-effect that we're not supporting symlinked appmap directories anymore. I don't think that it's used at all and `build/appmap` and `target/appmap` are never symlinks by default.


ahtrotta (approved):
Looks good and worked well when I manually tested it, thanks!

Issue #506, opened 2023-12-06
JetBrains sign in via license key


brikelly:
This task is to create an equivalent of [this PR for VS Code](https://github.com/getappmap/vscode-appland/pull/851).

The desired UX is as follows:

- Users who cannot use GitHub or GitLab to sign in can instead enter a license key. This key will be retrieved through other means, e.g. by contacting our support team. 
- After installing the AppMap plugin from the marketplace, a new menu option will be available under `Tools` -> `AppMap` called `Enter license key`. This can be placed at the bottom, in the third section of items:
![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/573108/c4b671f6-be1e-4e98-95b8-0d79e46e5a97)
- Once the user clicks on this menu item, they are presented with a modal dialog box with title "Enter your AppMap license key" and with a single text entry box named "license key". If they click the "OK" button then the auth checks occur. 
- The auth checks may require opening a browser similar to how the existing sign in is done. 
- If the key verification completes successfully, the "Sign In" panel should disappear and be replaced with the normal AppMap functionality, just as the existing GitHub/Lab flow does. A notification should also be displayed saying "Logged into AppMap".
- If the key verification fails, a notification should be displayed saying "License key verification failed".

Notes:
- If the user is already logged in with GitHub/Lab, entering a license key will replace their login with the new license key login instead. There is no need to warn the user that they're already logged in with another method.

PR #508, opened 2023-12-08
Action to sign in with a text API key

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/506

This PR adds a new action to manually sign into AppMap.
When the license key is modified in the text input box, a quick local validation is performed. It verifies that the base64-decoded key matches `user@host.com:a-license-id`. After confirming the dialog the key is validated with the remote AppLand host. 
The local validation is needed because the input dialog expects it to return quickly. The remote validation is not guaranteed to return quickly because it performing a remote HTTP request.

This PR adds tests. Testing remote validation is only possible with an invalid key because we can't store valid keys in git.

Initial view of input dialog:
image

With invalid key:
Screenshot_20231211_113030

After validation with the remote endpoint:
image

Notification after success:
image


ahtrotta (approved):

PR #530, opened 2024-01-15
Improve webview build and code sharing

jansorg:
- To remove deduplication, common code `messagePublisher.js` and `vsCodeBridge.js` is now shared between our different webviews
- Updates all webpack configs to avoid Vue development mode in our production builds and to fix local problems with the invocation of `yarn`
- Updates all AppMap components of all webviews to the latest versions


appland-findings/package.json
```diff
@@ -18,10 +18,9 @@
     "webpack-cli": "^4.6"
   },
   "dependencies": {
-    "@appland/components": "^3.5.0",
+    "@appland/components": "^3.25.1",
     "vue": "^2.7",
     "vue-template-compiler": "^2.7",
     "url": "^0.11"
-  },
-  "packageManager": "yarn@2.4.2"
```
ahtrotta, line 26: Why did you remove the `packageManager` field from the `package.json` files?

jansorg: Previously, I was able to run `yarn update`. I have yarn 1.22.21 installed.

But a few days ago things started to fail, as far as I remember node told me that `packageManager` is an indicator that `corePack` was used in the past and should be used now. After installing `corepack` it downloaded yarn2 (I think) and yarn2 did not create a node_modules directory and then webpack failed.
In summary: it's a mess and I could find a better fix.

What's the recommended way to stick to a specific package maanger or should we just use the default NPM?

ahtrotta: We don't have the `packageManager` field in our `package.json` files in `appmap-js` so I think it's fine.

ahtrotta (approved):

appland-findings/package.json
ahtrotta: Oh actually, we do have it in the root `package.json`. Let me look into this more.

ahtrotta: @jansorg it seems that suddenly these workspaces started using Yarn PnP instead of `node_modules`, so I added a `.yarnrc.yml` file to each workspace to enforce using `node_modules`. I believe this solved the problem, but can you verify that this solution works for you?

jansorg:
Thank you for the Yarn fix, @ahtrotta! It's working well now.

PR #532, opened 2024-01-19
Don't pollute project dir with log files of the AppMap Java agent

jansorg:
Our tests are executed with the AppMap Java agent attached.
By default, the agent is creating log files in the working directory, which is the project directory in our case.
We're now enforcing that logs are written into the build directory.


ahtrotta (dismissed):

ahtrotta:
Can you investigate why the build is failing?

jansorg:
@ahtrotta The log contained a network problem of the macOS runner, a restart fixed this.

ahtrotta (approved):

Issue #533, opened 2024-01-19
Implement Navie


dustinbyrne:
# Navie Explain

## Interactions
**Explain via refactorings**: In VS Code, a new refactoring action allows the user to select text, then `Explain with AppMap AI`. This action will pop open the web view, providing the selected text as the `question` prop (described below) to pre-fill the prompt input.

**Explain via command**: This opens the webview. In JetBrains, this might be an additional option under the AppMap main menu.

## Technical changes
**Launching `index`**:  
The index process should now launch with the additional argument `--port 0`. This requests the index process to listen for JSON RPC requests, with the OS allocating a port.

When the server becomes available, a message will be logged to stdout. Parse stdout for `/^Running JSON-RPC server on port: (\d+)$/` and record the port bound for that process.

Ensure the `APPMAP_API_KEY` environment variable is set to the user's API key. This is the same API key returned via authentication.

**Opening the Webview**:  
Instead of using the `Chat` page, `ChatSearch` is used.

The following props are needed:
`appmapRpcPort` - The port to connect to. This is derived from the regex above.
`apiKey` - The user's API key retrieved via existing authentication.
`savedFilters` - Saved filters shared with the AppMap view. These may change, and updates should be handled similar to the AppMap webview.

Optional:
`question` - Text to pre-load the input field with. In VS Code, this is used by the refactor action to populate the input with the selected text.

If the RPC port has not yet been detected, do not open the window. In VS Code, a notification appears that states:
> AppMap Explain is not ready yet. Please try again in a few seconds.

If multiple modules are open in the same window, a selector opens, requesting the user to select a module to communicate with. It's expected that each module already runs its own `index` process, so this step is to disambiguate which port to connect to.

`Content-Security-Policy` updated to include:
```
connect-src http://localhost:${rpcPort} https://api.getappmap.com
```

## VSCode References
- Backend initialization and message handling of the web view
  https://github.com/getappmap/vscode-appland/blob/develop/src/webviews/chatSearchWebview.ts
- Chat search webview front end
  https://github.com/getappmap/vscode-appland/blob/develop/web/src/chatSearchView.js
- Index process initialization and parsing
  https://github.com/getappmap/vscode-appland/blob/develop/src/services/indexProcessWatcher.ts

PR #543, opened 2024-01-24
Enable Navie in the user interface

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/533

This PR makes Navie available to users.

The AppMap CLI processes are restarted when the API key changes. The indexer JSON-RPC service needs an up-to-date API key to provide valid responses for the Navie webview.

If an editor is opened and text is selected, then this text is taken as default value for Navie's question text input.
If an editor is opened, then the editor's file is used as context to locate the matching appmap.yml file and the matching AppMap indexer JSON-RPC service.

Each execution of the "Explain with AppMap" action opens a new Navie webview.

Notification when Navie is not ready yet. We could change it to a modal message box instead of a notification:
image

Navie tab:
image

Popup if there's more than one AppMap indexer available:
Screenshot_20240125_124758


ahtrotta (changes requested):
There are some small hiccups that I'd like you to investigate. It's possible these are on our end, but I haven't seen them happen before. These were found when using `2023.1.1` with the `spring-petclinic` project.

When trying to use the select menu for AppMap search results, it doesn't always render correctly:

image

This also happens when using the select menu for selected code objects:

image

In the AppMap in the Navie webview, the `Dependency Map` is blank, and when the AppMap first loads, the zoom controls are rendered incorrectly:

image

jansorg:
@ahtrotta I've rebased on develop and also updated the navie components to the latest version.
I'm not sure what I could do about the rendering issues. Unless it's somehow caused by the plugin's index.html and/or the init of the JS application, it's all inside the webview and out of the plugin's control. It may be problems with the JCEF webkit integration, but that usually not fixable for plugins.

The dependency view and data is managed by the webview itself, AFAIK. The indexer's JSON-RPC service is most likely providing the data needed for the view. Or it there a property I need to pass to the navie webview?

ahtrotta:
> @ahtrotta I've rebased on develop and also updated the navie components to the latest version. I'm not sure what I could do about the rendering issues. Unless it's somehow caused by the plugin's index.html and/or the init of the JS application, it's all inside the webview and out of the plugin's control. It may be problems with the JCEF webkit integration, but that usually not fixable for plugins.

I haven't seen the select menu rendering issues happen since the first time, and it's not a blocking issue anyway, so we can call this resolved.

> The dependency view and data is managed by the webview itself, AFAIK. The indexer's JSON-RPC service is most likely providing the data needed for the view. Or it there a property I need to pass to the navie webview?

I managed to track this down to a missing import and I pushed a fix for it.

ahtrotta (changes requested):
This is working pretty well, thanks!

An issue that I found that needs to be resolved is when there are multiple `appmap.yml` files within a module then we need to have the user select which index process they want to use for `AppMap Explain`. For example, in the screenshot below, there are a total of three `appmap.yml` files, and there is an index process associated with each, but the user is only prompted with a menu showing two options. We need to have a menu with three options here:

image

For the relevant changes in VS Code, see [this commit](https://github.com/getappmap/vscode-appland/pull/872/commits/0f371fe3d6756d55725c57fa94547a548c64c337).

ahtrotta:
@jansorg since we just updated `@appland/components` again, can you rebase this?

ahtrotta:
I've found an issue. If I have multiple `appmap.yml` files in a module and then I open Navie, I get the expected selection menu:

image

However, if I close Navie, keep the same selection in the project file explorer menu (`MysqlTestApplication` in the screenshot), and then re-open Navie, then I don't get selection menu and it doesn't remember my previous choice.

We should make it so that every time Navie is opened the user should see the selection menu.

ahtrotta:
@jansorg I found another issue. If I delete an `appmap.yml` the indexer process is not terminated and when I open Navie, I still get the selection menu even if there is only one `appmap.yml` left in the project.

ahtrotta:
@jansorg I found a major problem. When using `2022.3`, it is not possible to close and re-open an AppMap multiple times. It will lead to a `404` error.

image

ahtrotta (approved):
This is working well, thanks!

The only small problem I'm noticing now is that when a user clicks the export button in a sequence diagram, they get this menu:

image


But we haven't implemented the `JSON` option yet. I don't think this should hold up this feature, but it's worth noting.

jansorg:
@ahtrotta Thanks for the review! Please feel free to merge this when you're happy with it.

Issue #535, opened 2024-01-20
Add compatibility with 2024.1 EAP


jansorg:
The first EAP is available now and the AppMap plugin should be compatible.

PR #553, opened 2024-02-02
Support JetBrains 2024.1 EAP releases

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/535

This adds compatibility with 2024.1 eap. Compilation did not show problems and all tests passed.
I'm waiting for 2024.1 eap3 to be fully published before making this PR available for review.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,10 +1,10 @@
 pluginVersion=0.50.0
 
 sinceBuild=213.0
-untilBuild=233.*
+untilBuild=241.*
 
 # run plugin verifier for the earliest and latest supported versions
-ideVersionVerifier=IC-2021.3.3,IC-2023.3.2
+ideVersionVerifier=IC-2021.3.3,IC-2023.3.2,IC-241.10840.26
 
 lombokVersion=1.18.24
 
@@ -16,6 +16,7 @@ ideVersion=IC-2021.3.3
 #ideVersion=IC-2023.1.1
 #ideVersion=IC-2023.2
 #ideVersion=IC-2023.3.2
+#ideVersion=IC-241.10840.26-EAP-SNAPSHOT
 
 org.gradle.jvmargs=-Dfile.encoding=UTF-8
 org.gradle.parallel=false


apotterri:
@jansorg What's the reason for having `ideVersion` and `ideVersionVerifier`? When are each of them used?

jansorg:
@apotterri `ideVersion` defines the SDK used for compilation and for tasks like `runIde`, CI is overriding it with `-PideVersion=...` to build and test against a specific version.
`ideVersionVerifier` is used for the JetBrains plugin verifier, which checks binary compatibility for each version contained in this property. Because it's a list we're not using `ideVersion` here.

apotterri:
Why does CI override it? When do you choose to change what's in `gradle.properties` and what's used by CI?

jansorg:
Please take a look at the GitHub Actions of this repo for the details.
CI is overriding the version for Windows because an earlier version is failing for unknown reasons on Windows with the plugins.

Compilation for releases needs to happen with the earliest version for binary compatibility. That's why it's enabled by default.
Later versions in gradle.properties are enabled locally during development to verify compilation and to test or reproduce a particular issue.

apotterri:
I have looked at the GH action. It's pretty thin on details, which is why I asked the question.... ;)

Ah, so `ideVersion` is serving double duty: it's used to specify binary compatibility, but it's also the version that's tested against (unless it's overriden, as for Windows). Would it make sense to add a matrix entry that tests a newer (maybe latest) `ideVersion`?

I'm asking because I added a test to the agent to ensure that it works with the plugin: https://github.com/getappmap/appmap-java/blob/65b816676a7052fdfa56b5d535a548d0fe107425/agent/test/intellij/intellij.bats#L35C17-L35C41 . It pins `ideVersion` to a version that failed with 1.26.2, because the oldest `ideVersion` didn't fail. I didn't feel completely comfortable with the pin, and I wondered if it was the right thing to do.

jansorg:
Glad to know why you're asking :smile: 

In general `ideVersion` defines the version of the SDK the plugin is built for. For releases we have to built against the earliest supported SDK, due to binary compatibility.
IMHO we have to run tests at least against the SDK used to compile the published plugin. That's what we're doing now.

We could certainly add a built-matrix, e.g. for the earliest and the latest version (tracking now at https://github.com/getappmap/appmap-intellij-plugin/issues/557).

For your commit to appmap-java: unless we could run appmap-intellij against a nightly build of appmap-java, I think that the test should remain in appmap-java.
Because the agent is working fine with the earlier JetBrains SDKs, you'll have to instruct gradle to use a different SDK. I don't think we should change the default in the plugin. 
Pinning the version is okay, but it could become outdated when we're dropping supported SDKs from the plugin. 

The possible ways to test with appmap-java I can think of:
1. Run plugin tests in a matrix against a nightly build on appmap-java
2. Run plugin tests in a matrix against a nightly, build upstream appmap-java and use it as agent
3. appmap-java runs tests against the JetBrains plugin, like your test

ahtrotta (approved):

Issue #558, opened 2024-02-05
Support nested AppMap directories


jansorg:
For this directory structure, it should launch three sets of AppMap CLI processes, not just two.
```text
dir1/appmap.yaml
dir1/subdir1/appmap.yml
dir2/appmap.yaml
```

PR #560, opened 2024-02-07
Jansorg/recursive cli

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/558


plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
```diff
@@ -345,23 +326,24 @@ protected void requestVirtualFileRefresh(@NotNull Path path) {
     }
 
     private void doRefreshForOpenProjectsLocked() {
-        var topLevelRoots = new VfsUtilCore.DistinctVFilesRootsCollection(VirtualFile.EMPTY_ARRAY);
+        var enabledRoots = VfsUtilCore.createCompactVirtualFileSet();
         for (var project : ProjectManager.getInstance().getOpenProjects()) {
-            if (!project.isDefault() && !project.isDisposed()) {
-                for (var contentRoot : ProjectRootManager.getInstance(project).getContentRoots()) {
```
jansorg, line 351: Scanning the content roots for `appmap.yml` isn't correct. This code was there since the initial revision of this file.
`appmap.yml` files should be detected everywhere, e.g. in subdirectories of content roots and not just as files inside a content root dir.
We're already querying the index in other places.

ahtrotta (approved):

PR #568, opened 2024-02-13
Workaround for exception "Read access is allowed from inside read action"

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/565
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/564

This is a workaround for the exceptions linked above.
As far as I can tell the AppMap plugin is calling `getJavaParameters()` correctly, because this method is handling its own `ReadAction`. The exceptions indicate that `getJavaParameters()` did not expect that a ReadAction is required and therefore did not use it. The code called in the stacktraces still expect a `ReadAction` and we're now falling back to use it if the first call of `getJavaParameters()` caused an exception.

I'm unable to reproduce the exception locally, but I'm confident that this is fixing the problem.


ahtrotta:
@jansorg is this ready for review? We'd like to get this fix in when we release the next version of the JetBrains plugin.

jansorg:
It'll be ready soon, I'll ping you when it's done. I'm currently verifying the fix because I'm unable to reproduce the original exception (and thus can't write a test for this).

ahtrotta (approved):

Issue #566, opened 2024-02-13
Exception about slow operations in `AppMapGradleAgentRunner.verifyJdk`


jansorg:
I've seen this when I tested with 2023.2.
- 2023.2
- clone of spring-petclinic without an appmap.yml
- Import as Gradle project
- In the Gradle project settings, set Gradle JVM to "JAVA_HOME" 
- Execute a Gradle-based test run configuration

```
java.lang.Throwable: Slow operations are prohibited on EDT. See SlowOperations.assertSlowOperationsAreAllowed javadoc.
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:376)
	at com.intellij.util.SlowOperations.assertSlowOperationsAreAllowed(SlowOperations.java:101)
	at com.intellij.openapi.vfs.newvfs.persistent.FSRecordsImpl.update(FSRecordsImpl.java:718)
	at com.intellij.openapi.vfs.newvfs.persistent.PersistentFSImpl.findChildInfo(PersistentFSImpl.java:603)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findInPersistence(VirtualDirectoryImpl.java:154)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.doFindChild(VirtualDirectoryImpl.java:137)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:83)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:535)
	at com.intellij.openapi.vfs.newvfs.impl.VirtualDirectoryImpl.findChild(VirtualDirectoryImpl.java:51)
	at com.intellij.openapi.vfs.impl.FilePartNode.findChildThroughJar(FilePartNode.java:492)
	at com.intellij.openapi.vfs.impl.FilePartNodeRoot.findFileFromRoot(FilePartNodeRoot.java:313)
	at com.intellij.openapi.vfs.impl.FilePartNodeRoot.trieDescend(FilePartNodeRoot.java:218)
	at com.intellij.openapi.vfs.impl.FilePartNodeRoot.findOrCreateByPath(FilePartNodeRoot.java:184)
[...]

PR #569, opened 2024-02-13
Retrieve Gradle JDK in a background thread

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/566

The issue contains instructions how to reproduce.
The retrieval of the Gradle JDK became a "slow operation" in 2023.2 and we must not fetch it on the UI dispatch thread (i.e. the EDT).


ahtrotta (approved):
This fixes that issue for me, thanks!

Issue #572, opened 2024-02-15
Avoid using `org.jetbrains.annotations.ApiStatus.Internal`


ahtrotta:
The plugin uses API annotated with `org.jetbrains.annotations.ApiStatus.Internal`. Such APIs are private and cannot be used outside the IntelliJ Platform itself. To find a recommended replacement API, check [this article](https://plugins.jetbrains.com/docs/intellij/api-internal.html).

Issue #500, opened 2023-11-15
Usage of internal API should cause CI failure


jansorg:
@dividedmind had the idea to fail CI if internal API is used again.
We could implement this after https://github.com/getappmap/appmap-intellij-plugin/pull/498 was merged. The plugin verifier can be configured in a way to fail when usage of internal API was detected.

PR #573, opened 2024-02-16
Don't use internal API of 2024.1 eap

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/572
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/500

2024.1 eap4 declared a few more methods as internal API.
To detect this earlier next time, I changed the settings of the plugin verifier to terminate unsuccessfully if internal API is used.

`JavaVersion` is now internal: There's no non-internal replacement. We're copying the class into the plugin. Please note that intellij-community is Apache2 and the AppLand plugin is MIT. I added the NOTICE.txt file to our plugin zip as it's required by the Apache2 license. I also kept the Apache 2 line in the file, https://github.com/getappmap/appmap-intellij-plugin/pull/573/commits/520d9c840d2cd0fd573e29865553f9eccac6d89a#diff-49ed108efcb06c07d51c5957fe32303829b0fe9794de9df7ef680b588e60ebef

The other use of internal API could be fixed in a better way.

https://github.com/getappmap/appmap-intellij-plugin/pull/573/commits/c56086c2dcc4979983d1b061589ed5787a660125 moves two classes, which are only used in a Java project, into our Gradle module, which depends on the JetBrains Java plugin. This isn't strictly needed, but cleaner and I was hoping that the Java plugin provided a non-internal API for `JavaVersion`.

There's one remaining problem, which I reported to JetBrains at https://youtrack.jetbrains.com/issue/IDEA-346395/2024.1-eap-breaks-API-compatibility-of-OAuthCallbackHandlerBase
We're currently unable to fix this unless we created builds for each supporte major version of the SDK. I'm still hoping that JetBrains is going to fix this for 2024.1.
The plugin verifier did not detect this, so uploading should be okay. Users on 2024.1 will get an exception, though.


ahtrotta:
@jansorg thanks for addressing this, I'm a bit confused, though. I don't see `org.jetbrains.annotations.ApiStatus.Internal` removed anywhere in this PR. Where was it being used and how does this fix it? 

Also, the automated upload worked fine when I submitted `0.56.0`. I was informed via email that it didn't pass by a human reviewer.

jansorg:
> @jansorg thanks for addressing this, I'm a bit confused, though. I don't see `org.jetbrains.annotations.ApiStatus.Internal` removed anywhere in this PR. Where was it being used and how does this fix it?

The `Internal` API is not in our plugin. The problem was, that the AppMap plugin used API of the JetBrains SDK, which was marked as internal. The SDK's class `JavaVersion` was marked as internal api, so we stopped using it. Class `JavaVersion` was marked internal because the class's package was made internal. On JetBrains slack it was mentioned that this may be unintentional, but to release now we have to drop the use of JavaVersion.

The other use of internal API was the call of a method of the SDK's class `FileIndexFacade`, which we also replaced with something else.
 
> Also, the automated upload worked fine when I submitted `0.56.0`. I was informed via email that it didn't pass by a human reviewer.

The reviewer checked the results of the plugin verifier :) You can find them at https://plugins.jetbrains.com/plugin/16701-appmap/edit/versions/stable/487373, the first item (shown in red) in section `Compatibility verification`.

The verification happens after upload. You or someone else of the AppMap org should have received an email about the detected compatibility problems, though.

ahtrotta:
Thanks for the explanation, Joachim, that makes a lot of sense!

I was just testing this PR manually, and I ran into a problem. If I navigate away from an AppMap webview by opening another tab, and then I return to the previous tab, it is blank:

image

Can you look into this?

ahtrotta:
> Sure, but I don't think that this is related to this PR. I've seen this with older IDE versions, but not with the current version. IMO It's most likely caused by multiple JCEF editors opened at once. Which IDE version did you test with?

Interesting, I was testing on `2023.3.2`, so maybe it's something else?

jansorg:
Yes, that's entirely possible. 
Is this reliably showing up with this PR, but not on `develop`?

ahtrotta:
> Yes, that's entirely possible. Is this reliably showing up with this PR, but not on `develop`?

Now I can't reproduce it at all, so I'm not sure what's going on.

ahtrotta (approved):

Issue #586, opened 2024-02-21
Freeze during indexing


jansorg:
During indexing, the following threads blocked the UI:

```text
"AWT-EventQueue-0@3193" prio=6 tid=0x1e nid=NA waiting for monitor entry
  java.lang.Thread.State: BLOCKED
	 waiting for ApplicationImpl pooled thread 3@6536 to release lock on <0xb2c2> (a appland.cli.DefaultCommandLineService)
	  at appland.cli.DefaultCommandLineService.getActiveRoots(DefaultCommandLineService.java:142)
	  at appland.webviews.navie.NavieEditorProvider.chooseIndexerJsonRpcPort(NavieEditorProvider.java:108)
	  at appland.webviews.navie.NavieEditorProvider.openEditor(NavieEditorProvider.java:63)
	  at appland.actions.OpenAppMapNavieAction.actionPerformed(OpenAppMapNavieAction.java:18)
	  at com.intellij.openapi.actionSystem.ex.ActionUtil.lambda$performActionDumbAwareWithCallbacks$4(ActionUtil.java:235)
	  at com.intellij.openapi.actionSystem.ex.ActionUtil$$Lambda$1331.2053343644.run(Unknown Source:-1)
	  at com.intellij.openapi.actionSystem.ex.ActionUtil.performDumbAwareWithCallbacks(ActionUtil.java:256)
	  at com.intellij.openapi.actionSystem.ex.ActionUtil.performActionDumbAwareWithCallbacks(ActionUtil.java:235)
	  at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.lambda$performAction$0(ActionMenuItem.java:289)
	  at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter$$Lambda$4220.1650303647.run(Unknown Source:-1)
	  at com.intellij.openapi.wm.impl.FocusManagerImpl.runOnOwnContext(FocusManagerImpl.java:233)
	  at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.performAction(ActionMenuItem.java:281)
	  at com.intellij.openapi.actionSystem.impl.ActionMenuItem$ActionTransmitter.actionPerformed(ActionMenuItem.java:296)
	  at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1967)
[...]

PR #587, opened 2024-02-21
Possible freeze of UI if "Explain with Navie" is invoked during indexing

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/586

This fixes the deadlock I noticed while testing "Explain with Navie".
A background thread locked the active AppMap dirs and waited for indexing to complete, but the "Explain with Navie" action attempted to access the same information under a lock. This froze the UI thread while indexing was still in progress.

### DIFF
--- a/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
+++ b/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
@@ -59,7 +59,7 @@ public class DefaultCommandLineService implements AppLandCommandLineService {
 
     // toString() uses unguarded access, synchronizing it could create deadlocks
     @GuardedBy("this")
-    protected final Map<VirtualFile, CliProcesses> processes = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


ahtrotta:
When testing, I'm still running into the problem described above, I think. I tried to use Navie a couple of times and it would stop responding in the middle of a response:

image

I closed Navie, waited a while, then re-opened Navie and tried again and it worked. So it seems like there might be a problem with the indexer still.

jansorg:
@ahtrotta With "indexer", I meant the IDE's indexer of file content. In this case, the "Explain with Navie" did not complete and the IDE UI froze.

The issue you found with the AppMap indexer is new. Do you mind to create a new issue for this?

ahtrotta:
Now I can't reproduce that error. I'm starting to wonder if there's something wrong with my dev environment. I'll make an issue if I see that happen again.

ahtrotta (approved):

PR #589, opened 2024-02-21
Exception "java.lang.IllegalArgumentException: invalid hex byte"

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/579

The preview JetBrains Markdown is opening URLs, which don't work with `path()`.
Our request handler is only suitable for our own URLs, so we can safely skip URLs we're unable to parse.


ahtrotta:
How can I reproduce this error?

jansorg:
@ahtrotta I don't know. It was reported by a user and seems related to the preview of the JetBrains Markdown plugin (it's using the built-in HTTP server, too). But I don't know what exactly is triggering this bug.

ahtrotta (approved):

Issue #591, opened 2024-02-23
Instructions are updated to include Navie


dustinbyrne:
The instructions pages should be:
- Add AppMap to your project
- Record AppMaps
- Ask AppMap Navie AI

"Explore AppMaps" and "Runtime analysis" should be removed.

"Ask AppMap Navie AI" simply opens the Navie UI and checks off the step.

jansorg:
@dustinbyrne I must also update the install guide webview to the latest version, right?

The "Explore AppMaps" page is currently also shown by the "You've created your first AppMap" notification. What should button "View your AppMaps" of this notification show instead?

The plugin also tracked if the used opened page "Runtime Analysis" and passed this as `investigatedFindings` project metadata property to the install guide webview. I suppose that this should be dropped, too?

PR #592, opened 2024-02-27
Update Installation Guide

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/591

This updates the install instructions panel to only show three steps.
- It updated the install instructions webview to the latest version
- "Ask AppMap Navie AI" is checked if the users previously opened the Navie webview in the current project.
- Notification "You opened your first AppMap" is updated to open Navie when the user clicks the action link of the Notification

(AFAIK this should not be merged indo develop at this time)


ahtrotta (approved):

Issue #590, opened 2024-02-21
Handle `show-appmap-tree` event from Navie webview


ahtrotta:
For the equivalent PR in VS Code, see [here](https://github.com/getappmap/vscode-appland/pull/889).

Handle the `show-appmap-tree` event from the Navie webview introduced in `@appland/components` version `4.2.0` by showing and expanding the `AppMaps` section of the AppMap toolbar:

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/0c9fd24a-1924-429d-8825-f5961c6dd9fe)

PR #593, opened 2024-02-27
Handle show-appmap-tree event from Navie webview

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/590

Opens the AppMap tool window and sets focus on the list of AppMaps.
The other panels of the tool window are collapsed.

Because the AppMap tree component is nested into several other components, the logic to make it visible and focused is iterating on the hierarchy.

The updated Navie webview contains links with `target=_blank`. Previously, this opened an empty popup window inside the IDE. Now, these links are opened just like other external links. 
This change is applied to all our webviews.


ahtrotta:
@jansorg can you investigate the build failure?

jansorg:
@ahtrotta I checked the log of Travis CI. It's a flaky test failure, but I couldn't understand how this could happen with our code. Rerunning Travis CI "fixed" it.

ahtrotta (approved):

Issue #571, opened 2024-02-14
Notify the user to upgrade if they're on <= 2022 to use Navie


dustinbyrne:
Because input doesn't work in the web view, users must upgrade to 2023. A dialog box would be sufficient.

Show Navie, pop up dialog box.

PR #595, opened 2024-02-28
Notify user if Navie isn't fully functional

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/571

If the current IDE's webview JCEF isn't suitable for Navie, then show a notification to ask the user to update. If the user chose "don't show again", then the message is never shown again in any of the user's projects.

We're showing the message on top of the Navie editor tab, but because JCEF rendering is async the message is shown before the webview is fully visible.
If this is a problem, we could inject the message more deeply into the editor, e.g. when the webview's message that it's ready is received.

image

### DIFF
--- a/plugin-core/src/main/resources/messages/appland.properties
+++ b/plugin-core/src/main/resources/messages/appland.properties
@@ -197,6 +197,8 @@ webview.findingDetails.appMapNotFound.message=The AppMap file could not be found
 
 webview.navie.title=AppMap AI: Explain
 webview.navie.chooseAppMapModule.title=Choose AppMap module
+webview.navie.webviewBroken.title=AppMap Navie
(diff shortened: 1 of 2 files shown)


plugin-core/src/main/resources/messages/appland.properties
```diff
@@ -197,6 +197,8 @@ webview.findingDetails.appMapNotFound.message=The AppMap file could not be found
 
 webview.navie.title=AppMap AI: Explain
 webview.navie.chooseAppMapModule.title=Choose AppMap module
+webview.navie.webviewBroken.title=AppMap Navie
+webview.navie.webviewBroken.message=Your IDE has a known issue that typing into text input fields in the Navie is broken.\nPlease update to 2023.2 or later to fix this.
```
ahtrotta, line 201: Please change to:

Please update your IDE to version 2023.2 or later. Navie will not work with versions prior to 2023.2 because there is a known IDE issue that prevents typing into input fields.

PR #600, opened 2024-03-05
Fix compatibility with 2024.1 eap's backwards incompatible API of OAuthRequestHandlerBase

jansorg:
2024.1 is breaking the API we're using for "Sign in to AppMap" in an incompatible way. I've reported this at https://youtrack.jetbrains.com/issue/IDEA-346395, but so far JetBrains doesn't seem likely to fix this.

This PR is fixing the problem on our side to allow users of 2024.1 to sign in to AppMap.

This can be tested by using the "Sign in" action (at `Tools > AppMap`) with the current 2024.1 eap of a JetBrains IDE. It should also be tested with 2022.1, our earliest supported version.
I've successfully tested this with both versions.


ahtrotta (approved):
I logged in and out on `2024.1` EAP and on `2022.1.3` and it worked well for both.

PR #604, opened 2024-03-11
CI to automate updates of webview components

jansorg:


jansorg:
Merging because it's only a new CI workflow and now code changes. I can't test this workflow before it's on the main branch

Issue #603, opened 2024-03-08
Sign in is updated to support email authentication


dustinbyrne:
![image](https://github.com/getappmap/appmap-intellij-plugin/assets/8737782/c1ee3832-63ae-4f2b-8589-facb551e0d04)
![image](https://github.com/getappmap/appmap-intellij-plugin/assets/8737782/0e6b814f-0da7-4282-9a0f-a45a85dd7899)

A new sign in flow exists which allows users to provide their email address for a verification code. Upon entering the verification code, an API key is exchanged and the user is logged in.

### `activate`
The first (and only) argument passed along with this event is the users API key. Upon receiving this event, the API key should be used to authenticate the user and the sign in can be hidden.

PR #611, opened 2024-03-11
Support Email Sign-In

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/603

This PR updates the signin webview to the latest version and add the message-handler for the new "activate" message to support email-based signup.
This can only be tested manually, e.g. by signing-out first (`Tools > AppMap`) and then signing-in with the new workflow.

In older IDEs typing into text fields inside webview doesn't work. This is a problem for users using a broken IDE with email-based sign-in. 
- [ ] Should we display a similar message for such setups to tell the user that entering the email address into the text field does not work?


ahtrotta (approved):
This looks good, and let's merge this PR, but as you point out the email flow doesn't work for older IDEs. Is there any way that we can display a different sign-in webview depending on the version of the IDE?

jansorg:
@ahtrotta Sure, showing something else based on the IDE version is possible. Ideally, we'd use the same webview, but pass a special property to hide the email signup, for example. WDYT?

Issue #602, opened 2024-03-08
Navie view is up to date


dustinbyrne:
![image](https://github.com/getappmap/appmap-intellij-plugin/assets/8737782/54b024c4-7923-41e3-a91b-221a1c1733dd)

## New props
The most recent `ChatSearch` view introduces the following new props:

### `mostRecentAppMaps`
A list of AppMaps, ordered by creation time in descending order. Limited to 10 elements.  
This list should react to the creation of new AppMaps and automatically update the frontend. The VS Code implementation achieves this by [posting a message](https://github.com/getappmap/vscode-appland/blob/develop/src/webviews/chatSearchWebview.ts#L113-L116) to the webview upon change, and [updating data properties of the view upon receiving the update message](https://github.com/getappmap/vscode-appland/blob/develop/web/src/chatSearchView.js#L51-L59).
The shape of each element is as follows:
```ts
interface AppMapListItem {
  name: string, // The name of the item to be displayed. This is sourced from the AppMap data itself via `metadata.name`.
  recordingMethod: string, // Also sourced from the AppMap data itself, via metadata.recorder.type
  createdAt: string, // An ISO date string
  path: string, // The path on disk to this AppMap file. This path will be emit back to the plugin when the AppMap is opened.
}
```

### `appmapYmlPresent`
A `boolean` value indicating whether or not an `appmap.yml` exists for this module. Currently in VS Code this is always true. This flag will change soon as we're planning on letting users open Navie without an `appmap.yml` present.

## New events
The following new events are now emit from Navie and must be handled by the plugin backend.

### `open-appmap`
The first (and only) argument passed along with this event is the path of the AppMap. This is the same path defined in `AppMapListItem`.

### `open-install-instructions`
Similar to the existing `open-record-instructions`, this event should open the first step of the instructions (project picker) upon being received.

## Existing events
Both `show-appmap-tree` and `open-record-instructions` are still in use.

PR #612, opened 2024-03-11
Update Navie webview and show most recent AppMaps

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/602

This PR adds the new, required properties and message handlers

It listens for new AppMaps and updates the webview after AppMaps changed.
The list of most recent AppMaps is filtered by the selected AppMap directory.

Because there may be lots of AppMap updates, the update of the most recent AppMaps is debounced by 1s. This code also prevents updates after changes to AppMaps, but during indexing. The AppMap data is only reliably available after indeixng finished..


ahtrotta (approved):
Looks good and worked well when I tested it :+1:

PR #616, opened 2024-03-13
Drop Travis CI

jansorg:
As discussed with @dustinbyrne we're not using Travis for this project anymore.
We're still using GitHub Actions, of course.


ahtrotta (approved):
RIP Travis :skull:

Issue #621, opened 2024-03-19
Downloads respect JetBrains proxy settings


dustinbyrne:
If the JetBrains editor has been configured with proxy settings, all downloads should respect these settings. E.g.:
- AppMap JAR
- AppMap CLI binaries

jansorg:
Hm, this should be already the case. Do you have reports that it's not working?

PR #623, opened 2024-03-20
Tests to make sure that downloads use the IDE's proxy

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/621

Tests downloads of AppMap binaries and of the Java agent JAR.
The only change (except the new tests) is a change to the AppMap CLI URLs to unescape unnecessarily escaped `@` and `/` characters of the URL to make it work with the Mockserver.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -77,6 +77,11 @@ allprojects {
         if (project.name != "plugin-core") {
             testImplementation(project(":plugin-core", "testOutput"))
         }
+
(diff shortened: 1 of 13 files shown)


apotterri (approved):
Tests look good, thanks.

In general, it would be best to avoid the `chore` type in commit subject lines when a more specific type is available. For example, the commit here that adds tests would be better tagged as `test`. Thanks!

Issue #597, opened 2024-03-04
Verify that uninstalling the plugin is possible, including shutdown of AppMap processes


jansorg:
A user reported that uninstalling and unloading the plugin wasn't terminating.
https://x.com/alexandreverri/status/1762894679434047990?s=46&t=1DqxEuCOk31Q5QmqySaUOA

reedjones:
I have the same problem

jansorg:
@reedjones Thanks for the details. This isn't supposed to happen, of course.
Which OS and JetBrains IDE are you using?

jansorg:
Reproduced on Linux:
```
"AWT-EventQueue-0@3206" prio=6 tid=0x1e nid=NA waiting for monitor entry
  java.lang.Thread.State: BLOCKED
	 waiting for ApplicationImpl pooled thread 10@21030 to release lock on <0xc796> (a appland.cli.DefaultCommandLineService)
	  at appland.cli.DefaultCommandLineService.stopAll(DefaultCommandLineService.java:174)
	  at appland.cli.DefaultCommandLineService.stopAll(DefaultCommandLineService.java:167)
	  at appland.cli.DefaultCommandLineService.dispose(DefaultCommandLineService.java:249)
	  at com.intellij.openapi.util.ObjectTree.runWithTrace(ObjectTree.java:139)
	  at com.intellij.openapi.util.ObjectTree.executeAll(ObjectTree.java:169)
	  at com.intellij.openapi.util.Disposer.dispose(Disposer.java:219)
	  at com.intellij.openapi.util.Disposer.dispose(Disposer.java:207)
	  at com.intellij.serviceContainer.ComponentManagerImpl.unloadServices(ComponentManagerImpl.kt:958)
	  at com.intellij.openapi.client.ClientAwareComponentManager.unloadServices(ClientAwareComponentManager.kt:72)
	  at com.intellij.ide.plugins.DynamicPlugins.unloadModuleDescriptorNotRecursively(DynamicPlugins.kt:735)
	  at com.intellij.ide.plugins.DynamicPlugins.access$unloadModuleDescriptorNotRecursively(DynamicPlugins.kt:95)
	  at com.intellij.ide.plugins.DynamicPlugins$unloadPlugin$2.run(DynamicPlugins.kt:487)
	  at com.intellij.openapi.application.impl.ApplicationImpl.runWriteAction(ApplicationImpl.java:1015)
[...]

PR #628, opened 2024-03-20
Do not remove the process reference if it's not the terminated process

jansorg:
When I was debugging https://github.com/getappmap/appmap-intellij-plugin/issues/597 I noticed that processes were left running. This happened when I closed and reopened a project quickly.

Terminating AppMap processes takes a few seconds on this system. If there's already a new process for an AppMap directory when a process terminates and calls its termination listener, then the reference to the new process was removed and it was left running.
This is now fixed by this PR.

After the change CI become even more unstable. I've patched a few problems of process termination and the test setup code. This PR is required for https://github.com/getappmap/appmap-intellij-plugin/pull/679, because in that PR "Reopen project" is triggering the bug fixed by this PR.


jansorg:
Tests seem to be more unstable on GitHub with this PR, but are stable on the local system. I need to investigate  this before it's ready to merge.

ahtrotta (approved):

PR #630, opened 2024-03-20
Don't terminate AppMap processes on the EDT

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/597

#597 contains a stacktrace of a deadlock. `stopAll` was called on the EDT and waiting on the service lock and another thread, which locked the service, was waiting on the EDT.

Uninstalling and unlading a plugin for an update is also disposing the services of the JetBrains plugin, i.e. `dispose()` modified by this PR is called.

(Testing this fix for the deadlock is not implemented, because it's difficult and would require mocking a few services.)


jansorg:
For unknown reasons `appmap-analysis` is failing and breaking CI

ahtrotta (approved):

Issue #624, opened 2024-03-20
An initial agent JAR is bundled into the plugin


dustinbyrne:
- [ ] The latest agent JAR is bundled into the release
- [ ] Updates are still checked at run time

PR #632, opened 2024-03-21
Bundle the AppMap Java agent with plugin

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/624

Our Gradle setup was already downloading the latest version to run the tests with the AppMap Java agent.
We're reusing this to bundle the downloaded agent jar into the plugin.zip.

The bundled files is used as a fallback if the download of the remote agent JAR failed.


apotterri (approved):
LGTM

PR #633, opened 2024-03-21
Fix GitHub Actions

jansorg:
In other PRs the build was failing due to the plugin verifier reports.
By default, the build setup uses the latest version of the plugin verifier. Today a new version was released, which incorrectly flags our code.
I've reported this at
https://youtrack.jetbrains.com/issue/MP-6438/Plugin-verifier-1.365-incorrectly-flags-overridden-methods-of-OverrideOnly

This PR uses the previous version of the plugin verifier to make our build work again.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -193,6 +193,9 @@ allprojects {
             onlyIf { this.project == rootProject }
             mustRunAfter("check")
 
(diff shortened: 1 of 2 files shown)


apotterri (approved):
Thanks for this.

PR #639, opened 2024-03-25
Download the Java agent from Maven Central

### LINKED ISSUE #636: The Java agent JAR file is downloaded from Maven Central
The plugin should first try to download the latest version of the agent JAR from Maven Central. If it cannot. it should fall back to the current behavior of downloading from GitHub.

Failures should be written to the log. But, because the plugin now bundles a version of the agent JAR, they should just be warnings.

apotterri:
Fixes #636 .

### DIFF
--- a/plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
+++ b/plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
@@ -49,7 +49,8 @@ public final class AppMapJavaAgentDownloadService {
                 try {
                     downloadJavaAgentSync(indicator);
                 } catch (IOException e) {
-                    LOG.warn("failed to download AppMap Java agent", e);
+                    // User should have been warned already, so log at DEBUG
+                    LOG.debug(String.format("failed to download AppMap Java agent: %s (%s)", e, e.getCause()));
                 }
             }
         }.queue();
@@ -62,14 +63,8 @@ public final class AppMapJavaAgentDownloadService {
             return false;
         }
 
-        var latestAsset = GitHubRelease.getLatestRelease(indicator, "getappmap", "appmap-java")
-                .stream()
-                .filter(asset -> "application/java-archive".equals(asset.getContentType()))
-                .findFirst()
-                .orElse(null);
-
+        var latestAsset = getLatestAsset(indicator);
         if (latestAsset == null) {
-            LOG.warn("JAR assets not found in latest GitHub release of getappmap/appmap-java");
             return false;
         }
 
@@ -78,6 +73,7 @@ public final class AppMapJavaAgentDownloadService {
 
         // don't download if the asset file does already exist
         if (Files.exists(assetDownloadPath)) {
+            LOG.info(String.format("%s already exists, not downloading", assetDownloadPath));
             return false;
         }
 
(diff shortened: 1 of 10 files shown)


plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
```diff
@@ -102,6 +96,27 @@ public boolean downloadJavaAgentSync(@NotNull ProgressIndicator indicator) throw
         return true;
     }
 
+    @Nullable
+    private static Release.Asset getLatestAsset(@NotNull ProgressIndicator indicator) {
+        final Release[] releases = {getDefaultRelease(), GitHubRelease.INSTANCE};
```
jansorg, line 101: nit: Use `MavenRelease.INSTANCE` because the method mentions it's only for testing?

plugin-java/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
```diff
@@ -171,4 +186,9 @@ private static void downloadAgentJarFile(@NotNull ProgressIndicator indicator,
             return null;
         }
     }
+
+    // For testing only, don't depend on this otherwise.
+    static Release getDefaultRelease() {
```
jansorg, line 191: note: The SDK has a `@TestOnly` annotation to mark such cases.

plugin-java/src/test/java/appland/javaAgent/AppMapJavaAgentDownloadServiceProxyTest.java
```diff
@@ -39,16 +40,11 @@ public void httpProxyConnection() throws Throwable {
...
-
-        var hasJarDownloadRequest = Arrays.stream(proxyRequests).anyMatch(request -> {
-            var path = request.getPath().getValue();
-            return request.containsHeader("host", "github.com")
-                    && path.startsWith("/getappmap/appmap-java/releases/download/v")
-                    && path.endsWith(".jar");
-        });
-        Assert.assertTrue("The JAR of the AppMap agent must be downloaded using the configured HTTP proxy",
-                hasJarDownloadRequest);
+        mockServerRule.getClient().verify(
```
jansorg, line 43: Thanks, that's better!

plugin-java/src/test/java/appland/javaAgent/AppMapJavaAgentDownloadServiceTest.java
```diff
@@ -56,7 +56,7 @@ public void multipleDownloadRequests() throws IOException {
     public void lockFileGuard() throws IOException {
         var service = AppMapJavaAgentDownloadService.getInstance();
 
-        var jarAsset = GitHubRelease.getLatestRelease(new EmptyProgressIndicator(), "getappmap", "appmap-java")
+        var jarAsset = service.getDefaultRelease().getLatest(new EmptyProgressIndicator())
```
jansorg, line 59: With this change, is there still test coverage that the `getLatest()` is still working for GitHub?

jansorg (commented):
Thanks a lot!
I added a few notes and a question.

plugin-java/src/test/java/appland/javaAgent/AppMapJavaAgentDownloadServiceTest.java
apotterri: No, it doesn't. I've added another test: https://github.com/getappmap/appmap-intellij-plugin/pull/639/files#diff-59f897d536a484ce3373f5357e6a59093530b7aabbddf6a16de0ad9f792cb3d4R52

jansorg (approved):

PR #644, opened 2024-04-01
Add AppMap status report

### LINKED ISSUE #625: Add a status indicator to display the status of external resources
- Indicate the status of the JAR file (current version, up to date/failed to update) and where to find logs
- Indicate the status of the CLI binaries (current version, up to date/failed to update) and where to find logs

ahtrotta:
Fixes #625 

Allows users to check on the status of the AppMap tools. Specifically, the Java Agent, the CLI, and the Scanner. The included information about each tool is:

* Downloaded version
* Download location
* Latest version available for download

This does not include error messages related to tool download, nor does it include where the tool was downloaded from (i.e. Maven, GitHub, etc.). I think this is a good initial implementation and we can follow on with additional features.

There is a new dialog in the `Tools -> AppMap` menu: 

image

This will print a markdown file:

image


plugin-core/src/main/java/appland/actions/PluginStatus.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            @Override
+            public void onSuccess() {
+                if (statusReportFile != null) {
+                    var editorManager = FileEditorManager.getInstance(project);
+                    editorManager.openFile(statusReportFile, true);
+                }
+            }
+        }.queue();
```
jansorg, line 69: I don't think we should create a new file on disk.
It's possible to create an in-memory file, which is not stored on disk and dropped when the editor is closed.
See below :arrow_down: how this could be done.

Please move ` "Generating AppMap Plugin status report..."` into the message bundle `appland.properties` and use something like this to show it: `AppMapBundle.get("pluginStatus.progressTitle")`.

```suggestion
        new Task.Backgroundable(project, "Generating AppMap Plugin status report...", true, PerformInBackgroundOption.ALWAYS_BACKGROUND) {
            private volatile VirtualFile statusReportFile;

            @Override
            public void run(@NotNull ProgressIndicator indicator) {
                statusReportFile = new LightVirtualFile(STATUS_REPORT_FILENAME, statusReportText(indicator));
            }

            @Override
            public void onSuccess() {
                if (statusReportFile != null) {
                    var editorManager = FileEditorManager.getInstance(project);
                    editorManager.openFile(statusReportFile, true);
                }
            }
        }.queue();
```

plugin-core/src/main/java/appland/actions/PluginStatus.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Objects;
+
+
+public class PluginStatus extends AnAction implements UpdateInBackground, DumbAware {
+    private static final Logger LOG = Logger.getInstance(PluginStatus.class);
+    private static final String STATUS_REPORT_FILENAME = "appmap_status_report.md";
+
+    @Override
+    public void actionPerformed(@NotNull AnActionEvent event) {
+        createStatusReport(Objects.requireNonNull(event.getProject()));
```
jansorg, line 34: If there's no requirement to keep it static, I'd just inline `createStatusReport` here.

plugin-core/src/main/java/appland/actions/PluginStatus.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Nullable
+    private static String getBinaryReportHeader(CliTool type) {
+        if (type == CliTool.AppMap) {
+            return "### AppMap CLI Status\n\n";
+        } else if (type == CliTool.Scanner) {
+            return "### AppMap Scanner Status\n\n";
+        }
+
+        return null;
```
jansorg, line 112: Personally, I prefer `switch`:
```suggestion
switch (type) {
            case AppMap:
                return "### AppMap CLI Status\n\n";
            case Scanner:
                return "### AppMap Scanner Status\n\n";
            default:
                return null;
        }
```

plugin-core/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
```diff
@@ -98,28 +98,12 @@ public boolean downloadJavaAgentSync(@NotNull ProgressIndicator indicator) throw
...
-
-                if (latestAsset != null) {
-                    LOG.info("Got latest asset URL: " + latestAsset.getDownloadUrl());
-                    return latestAsset;
-                }
-            } catch (IOException e) {
-                LOG.info(String.format("Failed to query release, %s: %s", release, e));
-            }
-        }
+    public @NotNull Path agentDirPath() {
```
jansorg, line 101: Could you make this `getAgentDirPath()`, please?

plugin-core/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
```diff
@@ -98,28 +98,12 @@ public boolean downloadJavaAgentSync(@NotNull ProgressIndicator indicator) throw
...
+        return Paths.get(SystemProperties.getUserHome()).resolve(Paths.get(".appmap", "lib", "java"));
+    }
 
-        LOG.warn("Couldn't query any source for latest release.");
-        return null;
+    public @NotNull Path agentFilePath() {
```
jansorg, line 105: nit: `agentFilePath()` -> `getAgentFilePath()`

jansorg (changes requested):
Thanks! That's a very useful addition :)

If the CLI do not exist on disk, then this shown like this:
image

Perhaps add something like `<unavailable>`  or `-`?

I'd probably also put the latest version at the top, then `Your version`, then the download location because the location may wrap.

I added a few notes in the PR. Let me know if you need suggestions for the SDK...

plugin-core/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
ahtrotta: Done! :+1:

plugin-core/src/main/java/appland/javaAgent/AppMapJavaAgentDownloadService.java
ahtrotta: Done! :+1:

plugin-core/src/main/java/appland/actions/PluginStatus.java
ahtrotta: Yeah, that's nicer, I changed it to this

plugin-core/src/main/java/appland/actions/PluginStatus.java
ahtrotta: Thanks!

plugin-core/src/main/java/appland/actions/PluginStatus.java
ahtrotta: Done! :+1:

jansorg (approved):
Thanks, it's looking good now!
I've pushed https://github.com/getappmap/appmap-intellij-plugin/pull/644/commits/a18f02833d54933b733531d924d67c23d5e10e94 with a few minor changes, mostly to adapt to IntelliJ's code style settings.
Please feel free to merge if you're happy with the change.

ahtrotta:
> Thanks, it's looking good now! I've pushed [a18f028](https://github.com/getappmap/appmap-intellij-plugin/commit/a18f02833d54933b733531d924d67c23d5e10e94) with a few minor changes, mostly to adapt to IntelliJ's code style settings. Please feel free to merge if you're happy with the change.

Thanks! Looks good!

PR #648, opened 2024-04-04
feat: Allow configuring environment for AppMap services

### LINKED ISSUE #646: Bring your own key and LLM

dividedmind:
Fixes #646

Followups:
- documentation,
- restart services on config change,
- use credential storage for OpenAI key.

### DIFF
--- a/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
+++ b/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
@@ -476,18 +476,23 @@ public class DefaultCommandLineService implements AppLandCommandLineService {
                 || SystemInfo.isWindows && CpuArch.isIntel64();
     }
 
-    private static @NotNull KillableProcessHandler startProcess(@NotNull Path workingDir,
+    public static @NotNull KillableProcessHandler startProcess(@NotNull Path workingDir,
                                                                 @NotNull String... commandLine) throws ExecutionException {
+
+        final var settings = AppMapApplicationSettingsService.getInstance();
 
         if (!Files.isDirectory(workingDir)) {
             throw new IllegalStateException("Directory does not exist: " + workingDir);
         }
 
-        var command = new GeneralCommandLine(commandLine);
-        command.withWorkDirectory(workingDir.toString());
-        command.withParentEnvironmentType(GeneralCommandLine.ParentEnvironmentType.CONSOLE);
+        var command = new GeneralCommandLine(commandLine)
+                .withWorkDirectory(workingDir.toString())
+                .withParentEnvironmentType(settings.isCliPassParentEnv()
+                        ? GeneralCommandLine.ParentEnvironmentType.CONSOLE
+                        : GeneralCommandLine.ParentEnvironmentType.NONE)
+                .withEnvironment(settings.getCliEnvironment());
 
-        var apiKey = AppMapApplicationSettingsService.getInstance().getApiKey();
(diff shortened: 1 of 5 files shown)


plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
```diff
@@ -476,18 +476,23 @@ private static boolean isSupported() {
                 || SystemInfo.isWindows && CpuArch.isIntel64();
     }
 
-    private static @NotNull KillableProcessHandler startProcess(@NotNull Path workingDir,
+    public static @NotNull KillableProcessHandler startProcess(@NotNull Path workingDir,
```
jansorg, line 479: nit: it would be enough to make this package-visible to support testing

plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
```diff
@@ -476,18 +476,23 @@ private static boolean isSupported() {
...
                                                                 @NotNull String... commandLine) throws ExecutionException {
 
+        final var settings = AppMapApplicationSettingsService.getInstance();
```
jansorg, line 482: I usually don't make local variables `final` to improve readability, but it's a matter of taste. I'm fine either way.

plugin-core/src/main/java/appland/settings/AppMapApplicationSettings.java
```diff
@@ -31,6 +33,13 @@ public class AppMapApplicationSettings {
      */
     private volatile boolean showFirstAppMapNotification = false;
 
+    /**
+     * Map of environment variables to be set when starting AppMap services
+     * The key is the name of the environment variable and the value is the value of the environment variable.
+     */
+    private volatile Map<String, String> cliEnvironment = new HashMap<>();
```
jansorg, line 40: I'd prefer if you override the setter of this map to avoid re-assigning the field.
Currently, `apply` of the settings is re-assigning the map to an unknown map implementation, which may cause problems with serialization.
With our own setter we could always use map implementation it's considering safe.

If the setter is overridden, Kotlin will automatically use it.

```
public void setCliEnvironment(Map<String, String> cliEnvironment) {
        this.cliEnvironment = Map.copyOf(cliEnvironment); // unmodifiable map
    }
```

jansorg (changes requested):
Thanks, well done! :)

plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
dividedmind: Fair enough, I suppose it's a habit from JavaScript. I'll leave it as is.

plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
dividedmind: Thanks, my Java is a bit rusty!

plugin-core/src/main/java/appland/settings/AppMapApplicationSettings.java
dividedmind: Yup, good idea, thanks!

jansorg, line 40: nit: Oh, sorry, I missed this in the first pass. You could make this `Map.of()` to set an empty, unmodifiable map.

jansorg (approved):
Thanks!

PR #651, opened 2024-04-09
Use configured environment in the RPC service

dividedmind:


plugin-core/src/main/java/appland/settings/AppMapApplicationSettings.java
```diff
@@ -57,6 +60,21 @@ public void setCliEnvironment(@NotNull Map<String, String> environment) {
...
+        commandLine = commandLine.withParentEnvironmentType(cliPassParentEnv
+                ? GeneralCommandLine.ParentEnvironmentType.CONSOLE
+                : GeneralCommandLine.ParentEnvironmentType.NONE);
+        commandLine = commandLine.withEnvironment(cliEnvironment);
+        var key = apiKey;
+        if (isNotEmpty(key)) {
+            commandLine = commandLine.withEnvironment("APPMAP_API_KEY", key);
+        }
+        return commandLine;
+    }
```
jansorg, line 76: nit: Altough I don't like `final` on parameters or local variables for readability, reassigning parameters is often confusing and leads to bugs.

(I usually tag parameters with `@NotNull` as much as possible to help IntelliJ with its inspections.)

How about this instead?

```suggestion
public GeneralCommandLine applyServiceEnvironment(@NotNull GeneralCommandLine commandLine) {
        var key = apiKey;
        var environmentType = cliPassParentEnv
                ? GeneralCommandLine.ParentEnvironmentType.CONSOLE
                : GeneralCommandLine.ParentEnvironmentType.NONE;

        return commandLine.withParentEnvironmentType(environmentType)
                .withEnvironment(cliEnvironment)
                .withEnvironment(isNotEmpty(key) ? Map.of("APPMAP_API_KEY", key) : Map.of());
    }
```

plugin-core/src/test/java/appland/rpcService/DefaultAppLandJsonRpcServiceTest.java
```diff
@@ -2,18 +2,20 @@
...
 
 import java.util.Collection;
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.hasEntry;
```
jansorg, line 18: I didn't even know that we have Hamcrest in the classpath. I checked and it's a transitive dependency of a few of our other testing libraries (wiremock, for example). Using should be alright, I just haven't used it yet :)

jansorg (changes requested):
Thanks!
I added a minor note, a change would be great.

At first, putting command line handling into settings didn't seem right. But I couldn't think of a better, existing place. Let's put it into settings for now and extract it if more like this is added...

plugin-core/src/test/java/appland/rpcService/DefaultAppLandJsonRpcServiceTest.java
dividedmind: I didn't even know what Hamcrest was ;)
IntelliJ suggested this with AWS CodeWhisperer and it looked like exactly what I needed.

dividedmind:
> At first, putting command line handling into settings didn't seem right. But I couldn't think of a better, existing place. Let's put it into settings for now and extract it if more like this is added...

Same here, I spent a couple minutes considering other options, but I think I've seen a similar pattern somewhere in IntelliJ so I just went with it.

jansorg (approved):

PR #661, opened 2024-04-11
Drop support for 2022.1 up to 2023.1

### LINKED ISSUE #655: Drop support for 2022.* and 2023.1
If possible, we should drop support for 2022.1.
Support the last 4 releases should be more than enough, most developers update quickly.

By dropping support we have fewer things to test, fewer JCEF problems and we could drop workarounds.

jansorg:
Closes #655 

Drops supports for the older IDEs and replaces deprecated API with the now available alternatives.
For 2023.1 (the earliest supported version) the plugin verifier is now only reporting one use of deprecated. We're unable to drop it because we're supporting clickable links in notifications and there's no other way to achieve this.

Questions / open:
- [ ] I was unable to test "Sign in with Key" because I don't have a valid key at hand

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # AppMap for JetBrains IDEs
 
 ## System Requirements
-**IntelliJ IDEA 2022.1 or later** is required to use this plugin.
+**IntelliJ IDEA 2023.2 or later** is required to use this plugin.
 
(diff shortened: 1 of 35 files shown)


ahtrotta (approved):
The code looks good and it worked well when I tried it manually, thanks!

Issue #659, opened 2024-04-10
Respect top-level appmap.yml file in a multi-module Maven project


jansorg:
Use and detect appmap.yml at the project's root instead of creating a new appmap.yml file when tests are executed for a Maven sub-project.
Tested with https://github.com/google/gson

Related to https://github.com/getappmap/appmap-intellij-plugin/issues/546

PR #662, opened 2024-04-15
Improve support for multi-module Maven projects

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/659
This PR is based on https://github.com/getappmap/appmap-intellij-plugin/pull/661, which must be merged first.

This PR improves the support for projects with multiple modules.

Sample project structure:
- `/project`: Root directory of the project and content root of a module
- `/project/appmap.yml`: Existing AppMap config file
- `/project/submodule`: Content root of module `submodule`

Starting (maven) tests inside `submodule`:
Without this PR, a new config file was created at `/project/submodule/appmap.yml` instead of reusing the existing `/project/appmap.yml` file.

This PR fixes the search so that the top-level `appmap.yml` file is used instead and no new appmap configuration file is created. 
Only directories, which are included in a content root of the project are considered. 
Allowing `appmap.yml` files located outside the project would create a lot of problems, e.g. undetected changes to `appmap.yml` file or undetected changes to `.appmap.json` files.

Other changes:
- Property `appmap_dir` of an existing `appmap.yml` is only updated to `tmp/appmap` if it's missing. Before, it was always overridden, which could have set a wrong path for some edge-cases.
- New tests to verify the lookup of the appmap.yml file for nested modules
- Don't override an existing `-Dargline=...` but append out arguments instead.
- Add a new IDE registry key to enable the debug log of the agent. By default, the agent does not create log files anymore. If this registry key is enabled, log files are created and allow debugging issues with the Java agent on a user's system.

### DIFF
--- a/plugin-core/src/main/java/appland/files/AppMapFiles.java
+++ b/plugin-core/src/main/java/appland/files/AppMapFiles.java
@@ -80,6 +80,33 @@ public final class AppMapFiles {
     }
 
     /**
+     * Locate an existing appmap.yml file, searching in context and all its parent directories.
+     * This method does not use indexes, but needs to be invoked in a {@link ReadAction}.
+     * Only files on the local filesystem are considered.
+     *
+     * @param context     The file or directory to search from
+     * @param searchScope Scope to restrict the search for the appmap configuration file
+     * @return The appmap.yml file in the directory or its ancestors, if available
+     */
+    @RequiresReadLock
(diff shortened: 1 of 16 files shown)


ahtrotta:
@jansorg it looks like this has conflicts with `develop`, could you address those? Thanks!

jansorg:
@ahtrotta I've rebased this PR

ahtrotta (approved):

Issue #668, opened 2024-04-17
Update AppMap toolbar


ahtrotta:
## Changes

* Add a `NAVIE` section to the sidebar containing a large blue button that opens the Navie chat interface. The `NAVIE` section should be at the top.
* Rename `INSTRUCTIONS` to `APPMAP RECORDING INSTRUCTIONS`.
* Rename `APPMAPS` to `APPMAP DATA`.
* Change the text and button when there is no AppMap data (see image).
* Make the instructions and runtime analysis sections collapsed by default.

## Before

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/1971bcd4-ca51-48a7-b787-8c27d0526526)

## After

![Image](https://github.com/getappmap/appmap-intellij-plugin/assets/45714532/df097875-5f95-47df-9e97-1f9f8c077b4f)

Issue #667, opened 2024-04-17
RPC service receives workspace configuration


dustinbyrne:
A `v2.configuration.set` RPC method exists which ingests the following data:
```json5
{
  // This is the same value used today in v1.configuration.set
  "appmapConfigFiles": [
    '/home/db/dev/getappmap/vscode-appland/appmap.yml',
    '/home/db/dev/getappmap/appmap-intellij-plugin/appmap.yml'
  ],

 // This new property is an array of absolute paths to all projects open in the IDE 
  "projectDirectories": [
    '/home/db/dev/getappmap/vscode-appland',
    '/home/db/dev/getappmap/appmap-intellij-plugin',
    '/home/db/dev/bevyengine/bevy'
  ]
}
```

To upgrade, the following steps should be taken:
- The RPC method is updated to use `v2.configuration.set`
- A `projectDirectories` property is sent, where the value is an array of strings, and each string is an absolute path to a project directory open in the IDE.
- If `v2.configuration.set` fails, `v1.configuration.set` is used as a fallback.

dustinbyrne:
This depends on the release of https://github.com/getappmap/appmap-js/pull/1723

Pre-release binaries are available via https://github.com/getappmap/appmap-js/actions/runs/8723246396

PR #672, opened 2024-04-22
Add Navie section to the AppMap toolwindow

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/668

Questions:
- In a project without AppMaps, a click on "Open instructions" opens the installation guide. Is this still correct?

I've tested this with Light and Dark themes of 2022.1 and briefly with 2024.1.

All panels expanded:
image
image

Default state:
image
image

Project without AppMaps:
image


ahtrotta (changes requested):
Looks good! There are a few small changes that are needed:

* There's a small spelling mistake where AppMap is Appmap (see comment).
* The `AppMap Data` tool window should be expanded by default, not collapsed.
* Please make the `Create AppMap Data` a blue button in the same style as the `New Navie Chat` button.

Thanks!

jansorg:
@ahtrotta I've applied the requested changes except for the button.
I've also updated the Navie webvomponent because it was still displaying the red warning about missing AppMaps.
Implementing the button isn't straight-forward, unfortunately. There's not built-in support for it and I'll have to implement a custom Swing component to make this work. I was planning to implement #667 before adding the button.

ahtrotta (approved):

PR #673, opened 2024-04-23
Send project content roots to local AppLand JSON-RPC service

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/667

This PR implements JSON-RPC message `v2.configuration.set`. 
It send the project's content roots as property `projectDirectories`. Nested content roots are removed from the list sent to the JSON-RPC service.

For spring-petclinic the following message is generated:
```json
{
  "jsonrpc": "2.0",
  "method": "v2.configuration.set",
  "params": {
    "projectDirectories": [
      "/sources/spring-petclinic-2023-08-31"
    ],
    "appmapConfigFiles": [
      "/sources/spring-petclinic-2023-08-31/appmap.yml"
    ]
  }
}
```

Only appmap-analysis is failing on CI, I don't know what's causing this failure.

### DIFF
--- a/plugin-core/src/main/java/appland/rpcService/AppLandJsonRpcListener.java
+++ b/plugin-core/src/main/java/appland/rpcService/AppLandJsonRpcListener.java
@@ -2,6 +2,7 @@ package appland.rpcService;
 
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.messages.Topic;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.Collection;
 
@@ -14,5 +15,6 @@ public interface AppLandJsonRpcListener {
 
     void serverRestarted();
 
-    void serverConfigurationUpdated(Collection<VirtualFile> appMapConfigFiles);
+    void serverConfigurationUpdated(@NotNull Collection<VirtualFile> contentRoots,
+                                    @NotNull Collection<VirtualFile> appMapConfigFiles);
 }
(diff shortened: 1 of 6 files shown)


ahtrotta (changes requested):
When I run this in a project with AppMaps, I still see `No AppMap data available` and I don't see any context and Navie answers as if there isn't any context. I verified that this doesn't happen on `develop`.

image

ahtrotta:
Also, if you update `@appland/components` to `4.16.1` in `appland-navie`, then you'll be able to see the context in the right-hand side panel in the Navie chat interface.

jansorg:
I've updated to 4.16.1.
The latest JSON-RPC binary doesn't seem to support v2 yet, but still replies with a 204 status. I can't detect a bad request this way.

jansorg:
I've fixed the fallback handling, which is now used if the response `status != 2xx` or if the JSON-RPC response contains an error.
I've successfully tested the fallback with [@appland/appmap-v3.136.0](https://github.com/getappmap/appmap-js/releases/tag/%40appland%2Fappmap-v3.136.0).

dustinbyrne (approved):
The change looks good. I had to run/restart IntelliJ three times for the context to appear to be working though. I'm not sure what that was about. Maybe the CLI hadn't updated?

In any case I was ultimately able to request the configuration from the RPC server launched by the plugin and I can confirm it's getting configured properly.

jansorg:
> The change looks good. I had to run/restart IntelliJ three times for the context to appear to be working though. I'm not sure what that was about. Maybe the CLI hadn't updated?

It's possible that the CLI wasn't updated yet. The download progress should have been shown in the status bar and the JSON-RPC server would relaunch after the startup.

PR #674, opened 2024-04-23
Fix unstable tests

jansorg:
This is hopefully fixing all our unstable tests.

With this PR, we're moving away from light tests, because they're reusing project data between tests. The past has shown that this reuse is making our tests (more) unstable.
For example, services were reused between tests. This made it possible that the state of AppMap CLI processes was leaked into other tests, which expect a clean state.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,8 @@ jobs:
     test:
         strategy:
             matrix:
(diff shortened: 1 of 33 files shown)


jansorg:
Supersesed by https://github.com/getappmap/appmap-intellij-plugin/pull/762
If tests are still unstable, we probably need to implement the new test setup proposed by this PR. But atm it's too many changes for uncertain results.

Issue #677, opened 2024-04-25
Integrate the Navie context panel


dustinbyrne:
- `@appland/components` is updated in the Navie view
  - A new event is emitted, `open-location`, which indicates the user has requested to open contextual information in the editor, directly from the source.
  - The first (and only) argument passed along with this event is the location. This is a path (either relative or absolute), with an optional starting line number and optional ending line number in the following format: `path:startingLine-endingLine`. Below are all valid examples:
  ```
  C:\\Users\\example.appmap.json:35
  C:\\Users\\Default\\MyProject\\example.rb:35-99
  MyProject/example.rb:35-99
  MyProject/example.appmap.json
  /home/user/dev/appmap/example.java:56
  ```
  - If the path is an `appmap.json`, the starting line is actually an event ID. The [AppMap should be opened with the following state](https://github.com/getappmap/appmap-intellij-plugin/blob/develop/appland/webview.js#L34), where `${startLine} is substituted with the starting line or event ID.
   ```json
   {
      "currentView": "viewSequence",
      "selectedObject": "event:${startLine}"
    }
    ```
  - Otherwise, the location should be opened as a text document.
  - If a start and end line are provided, once open, the document should scroll to the range of text provided.
  - If only a start line is provided, the document should scroll to the start line
  - If neither are provided, the document simply opens and no scrolling takes place.
- ~Keep in mind, if the location is a relative path, it may match more than one module. For now, simply select the first match. We'll fix this in a future revision.~
  - A `directory` property will indicate where the file is located. It may still need to be searched for (i.e., a partial path), though the `directory` will identify the correct root directory. See [this source](https://github.com/getappmap/appmap-js/blob/a51cb16f8f81a240d7662ff23ff2a175d1a8c39b/packages/navie/src/context.ts#L48-L54) for more information. 

Sample context data is available here:
https://raw.githubusercontent.com/getappmap/appmap-js/main/packages/components/src/stories/data/navie_context.json

jansorg:
@dustinbyrne Are lines 1-based or 0-based?

jansorg:
@dustinbyrne Is VSCode displaying an error if no file for a given relative path was found?

kgilpin:
Context item should have a directory; this may be something we can fix now.

The file may still need to be located though, because the Java agent can’t
always determine full file paths.

dustinbyrne:
> @dustinbyrne Are lines 1-based or 0-based? Update: 1-based, it seems.

[...]

PR #678, opened 2024-04-26
Support `open-location` for Navie webview

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/677

This PR adds support for Navie's `open-location`.
Numbers of `name:line-line` are taken as 1-based and are adjusted to the IDE's 0-based numbers.
Relative paths are looked up in the available project files to find the best match.

I tested opening regular source files and also opening an AppMap using pre-release https://github.com/getappmap/appmap-js/releases/tag/%40appland%2Fappmap-v3.137.0

This PR is supporting both variants of the `open-location` message: `open-location "path/to/file"` and `open-location "path/to/file", "/root/dir"`. I wasn't able to test the 2nd variant, but AFAIK it would need a new release with an updated Navie webview, anyway.


appland-navie/webview.js
```diff
@@ -62,6 +62,7 @@ export function mountWebview() {
     });
 
     app.$on('open-install-instructions', () => vscode.postMessage({command: 'open-install-instructions'}))
+    app.$on('open-location', (path) => vscode.postMessage({command: 'open-location', path}))
```
kgilpin, line 65: Note than in an update coming soon, the enclosing directory of the path will also be provided. So only that directory will need to be searched for the file. You could include the directory parameter in this PR and treat it as optional?

kgilpin: Just be aware that you can’t simply join the directory to the path to identity the file. The file path location within the directory may not be known exactly (eg Jvm recordings).

kgilpin: You can see the WIP change here - https://github.com/getappmap/appmap-js/pull/1752/files

ahtrotta (approved):
Looks good and worked well when I tested it manually, thanks!

Issue #675, opened 2024-04-25
Bring your own OpenAI API key


dustinbyrne:
Users should be able to input their own OpenAI API key for use with Navie. When set, the RPC process should receive the environment variable `OPENAI_API_KEY` with this value.

Ideally, this value is not stored in plain text.

Upon changing the OpenAI API key, the user is prompted to reload the editor. If possible, the user can click to reload.

If the value is set to an empty string, it is cleared and `OPENAI_API_KEY` is no longer propagated. This value should take precedence over the environment variable settings.

Example flow in Visual Studio Code
[Peek 2024-04-25 12-57.webm](https://github.com/getappmap/appmap-intellij-plugin/assets/8737782/93bc6b96-7034-4a8d-8fef-b3d8c60b59d2)

PR #679, opened 2024-04-27
Allow users to specify their own OpenAI key in the settings

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/675

This PR adds a new setting to configure a custom OpenAI key.
There's currently no additonal action (e.g. `AppMap: Set OpenAPI key`), but it could be added if it's needed.

Settings are persisted in the IDE's password safe.

When the OpenAI key changed, a notification is displayed in each open project to allow the user to reload it. 
The application-wide JSON-RPC process **is NOT restarted** immediately after the change of the OpenAI key. A restart would invalidate the port used by open Navie webviews.

The notification is sticky, which means that it won't hide automatically. Otherwise the user might miss the required reload.

This PR is moving `applyServiceEnvironment` from the application settings class to the command line handler class, because it's now using both ApplicationSettings and the new `SecureApplicationSettings`.

https://github.com/getappmap/appmap-intellij-plugin/pull/628 is needed, because "Reopen project" added by this PR is triggering the bug fixed by #628 .

**Testing**:
- I've tested this on Linux and macOS Sonoma with the secure storage of the OpenAI key. I haven't seen any weird prompts or behavior.

Updated settings UI:
image

Notification:
image

### DIFF
--- a/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
+++ b/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
@@ -5,6 +5,7 @@ import appland.config.AppMapConfigFileListener;
 import appland.files.AppMapFiles;
 import appland.files.AppMapVfsUtils;
 import appland.settings.AppMapApplicationSettingsService;
+import appland.settings.AppMapSecureApplicationSettingsService;
 import com.intellij.execution.ExecutionException;
 import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.configurations.PtyCommandLine;
@@ -39,9 +40,12 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
+
+import static com.intellij.openapi.util.text.StringUtil.isNotEmpty;
 
 public class DefaultCommandLineService implements AppLandCommandLineService {
     private static final Logger LOG = Logger.getInstance(DefaultCommandLineService.class);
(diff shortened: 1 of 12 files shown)


kgilpin:
If the RPC process is restarted, then why does the user need to restart JetBrains?

kgilpin:
> macOS because the OS's password access confirmation was blocking IDE startup.

do you have more information on this? We are using a facility provided by JetBrains and it results in the IDE being blocked from starting?

jansorg:
> If the RPC process is restarted, then why does the user need to restart JetBrains?

I understood that all webview editors and also the JSON-RPC process need to be restarted. If it's only the process, then I could implement it that way.

kgilpin:
> > If the RPC process is restarted, then why does the user need to restart JetBrains?
> 
> I understood that all webview editors and also the JSON-RPC process need to be restarted. If it's only the process, then I could implement it that way.

You’re right because the RPC ports are out of date and I don’t think they can be updated. 

We can work on that in a future version.

jansorg:
> > > macOS because the OS's password access confirmation was blocking IDE startup.
> > 
> > 
> > do you have more information on this? We are using a facility provided by JetBrains and it results in the IDE being blocked from starting?
> 
> I'd need to test this with the current IDEs. I think it was caused by requesting secure storage during startup, but I'm not sure if that's still happening.

I tested on macOS Sonoma and it worked well.

jansorg:
I've made a few changes to the first version of this PR:
- after a change of the key, the JSON-RPC is NOT restarted. It would invalidate the key
- the notification does not contain an action to reload the project

ahtrotta:
@jansorg can you look into the build failure?

jansorg:
@ahtrotta It was a flaky test. I restarted the tests.
The unstable test which will hopefully and finally be fixed by https://github.com/getappmap/appmap-intellij-plugin/pull/674.

ahtrotta (approved):
Looks good and worked well when I tested it manually, thanks!

PR #691, opened 2024-05-02
fix: set language property in default config

### LINKED ISSUE #689: appmap.yml should include `language`
The default config generated by the plugin is missing the `language` entry. It should be set to one of `java`, `javascript`, `ruby`, or `python`, as appropriate.

apotterri:
When generating the default config file, set the language property.

Fixes #689 .

### DIFF
--- a/plugin-core/src/main/java/appland/config/AppMapConfigFile.java
+++ b/plugin-core/src/main/java/appland/config/AppMapConfigFile.java
@@ -31,11 +31,14 @@ public class AppMapConfigFile {
     private static final String NAME_PROPERTY = "name";
     private static final String APPMAP_DIR_PROPERTY = "appmap_dir";
     private static final String PACKAGES_PROPERTY = "packages";
+    private static final String LANGUAGE_PROPERTY = "language";
+    private static final String LANGUAGE_VALUE = "java";
 
     private final @NotNull ObjectNode yamlData;
 
     public AppMapConfigFile() {
         this(new ObjectNode(JsonNodeFactory.instance));
+        yamlData.set(LANGUAGE_PROPERTY, new TextNode(LANGUAGE_VALUE));
     }
 
     public @Nullable String getName() {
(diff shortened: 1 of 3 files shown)


plugin-core/src/main/java/appland/config/AppMapConfigFile.java
```diff
@@ -31,11 +31,14 @@ public class AppMapConfigFile {
...
     private static final String APPMAP_DIR_PROPERTY = "appmap_dir";
     private static final String PACKAGES_PROPERTY = "packages";
+    private static final String LANGUAGE_PROPERTY = "language";
+    private static final String LANGUAGE_VALUE = "java";
 
     private final @NotNull ObjectNode yamlData;
 
     public AppMapConfigFile() {
         this(new ObjectNode(JsonNodeFactory.instance));
+        yamlData.set(LANGUAGE_PROPERTY, new TextNode(LANGUAGE_VALUE));
```
jansorg, line 41: For consistency, I'd have put this line into `AppMapJavaPackageConfig`, but if you don't want to allow a missing language by default, then the constructor here is probably the better location.
https://github.com/getappmap/appmap-intellij-plugin/blob/3ff6d6f98bc432e6846ecff32627d81b12f715db/plugin-java/src/main/java/appland/execution/AppMapJavaPackageConfig.java#L117-L121

jansorg (approved):
Thanks!
How is VSCode treating multi-module projects with mixed languages, e.g. with a Java backend and a JS frontend module?
The appmap.yml is only created when the "Run with AppMap" button is clicked, so Java should be fine as a default. I just don't know what will happen with AppMaps in a submodule with JS, for example.

plugin-core/src/main/java/appland/config/AppMapConfigFile.java
apotterri: Yes, I put it here deliberately, because it should never be unset.

apotterri:
> Thanks! How is VSCode treating multi-module projects with mixed languages, e.g. with a Java backend and a JS frontend module? The appmap.yml is only created when the "Run with AppMap" button is clicked, so Java should be fine as a default. I just don't know what will happen with AppMaps in a submodule with JS, for example.

VS Code relies on the installer to create `appmap.yml`. If it can't determine the project type, it asks the user. This seems equivalent to deciding that the project is a Java project when the user chooses "Start with AppMap" in IntelliJ.

PR #694, opened 2024-05-03
Add keyboard shortcut open the Navie webview

jansorg:
This PR adds a shortcut for "Explain with Navie":
- Linux/Windows: `ALT + A, N`
- macOS: `Option + A, N`

AFAIK the VSCode plugin uses `Ctrl + A, N` on macOS. But `Ctrl+A` is already used in JetBrains IDEs. `Ctrl + A` is already assigned on macOS (action "go to start of line") and this action overrides our own shortcut.
`Option + A, N` is working fine on my setup of macOS.

### DIFF
--- a/plugin-core/src/main/resources/META-INF/appmap-core.xml
+++ b/plugin-core/src/main/resources/META-INF/appmap-core.xml
@@ -160,7 +160,12 @@
 
         <action id="appmap.generateOpenAPI" class="appland.actions.GenerateOpenApiAction"/>
 
-        <action id="appmap.openNavie" class="appland.actions.OpenAppMapNavieAction"/>
+        <action id="appmap.openNavie" class="appland.actions.OpenAppMapNavieAction">
+            <keyboard-shortcut keymap="$default"
+                               first-keystroke="alt A" second-keystroke="N"/>
+            <keyboard-shortcut replace-all="true"
+                               keymap="Mac OS X" first-keystroke="meta A" second-keystroke="N"/>
+        </action>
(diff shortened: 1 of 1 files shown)


dividedmind:
We've had a discussion about the keybindings and settled on Alt+A to start an Appmap chord, so then Alt+A N can be Navie, Alt+A E can be explain, etc. What do you think? I think IntelliJ support chords.

kgilpin:
Here's a suggestion as well:

Screen Shot 2024-05-03 at 10 40 48 AM

jansorg:
> We've had a discussion about the keybindings and settled on Alt+A to start an Appmap chord, so then Alt+A N can be Navie, Alt+A E can be explain, etc. What do you think? I think IntelliJ support chords.

@dividedmind What's the difference between "Navie" and "Explain"?

dividedmind:
> > We've had a discussion about the keybindings and settled on Alt+A to start an Appmap chord, so then Alt+A N can be Navie, Alt+A E can be explain, etc. What do you think? I think IntelliJ support chords.
> 
> @dividedmind What's the difference between "Navie" and "Explain"?

None really, currently, I think. I somehow had in my mind that there was a separate command to talk to Navie about current code selection. (Which could be a good idea, @kgilpin ?)

kgilpin:
For now there’s only one Navie command. If you’ve made a code selection, it will be applied.

dustinbyrne:
I'm going to close this until someone is able to dedicate time to identifying a worthwhile cross-platform keybinding.

PR #699, opened 2024-05-07
fix: only include source directories in config

### LINKED ISSUE #690: extra path entries in `appmap.yml`
Sometimes there are `path` entries added for non-application packages.

For example, the plugin generates this `appmap.yml` for waltz:
```
name: "waltz"
appmap_dir: "tmp/appmap"
packages:
- path: "app-group.add-aliases"
- path: "clients.c1.sc1"
- path: "licence.finos"
- path: "licence.spdx"
- path: "lineage"
- path: "liquibase"
- path: "measurable-ratings.edit-ratings"
- path: "org.finos.waltz.common"
- path: "org.finos.waltz.data"
- path: "org.finos.waltz.jobs"
- path: "org.finos.waltz.model"
- path: "org.finos.waltz.service"
- path: "org.finos.waltz.web"
- path: "report-grid.create-grid"
- path: "spark.template.freemarker"
- path: "static"
- path: "taxonomies"
- path: "user.registration"
```
It should really have no more than
```
- path: "org.finos.waltz.common"
- path: "org.finos.waltz.data"
- path: "org.finos.waltz.jobs"
- path: "org.finos.waltz.model"
- path: "org.finos.waltz.service"
- path: "org.finos.waltz.web"
```
and, ideally, even these would be collapsed to their common prefix, i.e. `- path: "org.finos.waltz"`.

apotterri:
When generating the default `appmap.yml`, make sure only source directories are included.

Fixes #690

### DIFF
--- a/plugin-java/src/main/java/appland/execution/AppMapJavaConfigUtil.java
+++ b/plugin-java/src/main/java/appland/execution/AppMapJavaConfigUtil.java
@@ -9,33 +9,16 @@ import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.concurrency.annotations.RequiresReadLock;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.jps.model.java.JavaModuleSourceRootTypes;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
+import java.util.Comparator;
+import java.util.stream.Collectors;
 
 public final class AppMapJavaConfigUtil {
     private AppMapJavaConfigUtil() {
-    }
-
-    /**
-     * Locate the directory, where AppMap JSON files should be stored in the given module.
-     * It always uses tmp/appmap as output directory. The returned tmp/appmap path is located in the most suitable
-     * content root of the module.
-     *
-     * @param module  Current module
-     * @param context Context to locate the most suitable content root for the AppMap output directory.
-     * @return The path to the tmp/appmap directory, where AppMap files should be stored.
-     */
-    @RequiresReadLock
-    public static @Nullable Path findAppMapOutputDirectory(@NotNull Module module, @NotNull VirtualFile context) {
-        ApplicationManager.getApplication().assertReadAccessAllowed();
-
-        var contentRoot = findBestAppMapContentRootDirectory(module, context);
-        var contentRootPath = contentRoot.getFileSystem().getNioPath(contentRoot);
-        return contentRootPath != null
-                ? contentRootPath.resolve(Paths.get("tmp", "appmap"))
-                : null;
     }
 
     /**
@@ -50,24 +33,16 @@ public final class AppMapJavaConfigUtil {
                                                                           @NotNull VirtualFile context) {
         ApplicationManager.getApplication().assertReadAccessAllowed();
 
-        // content root of the module, which contains the context file
-        var matchingModuleRoot = Arrays.stream(ModuleRootManager.getInstance(module).getContentRoots())
-                .filter(root -> VfsUtilCore.isAncestor(root, context, false))
-                .findFirst()
-                .orElse(null);
-        if (matchingModuleRoot != null) {
-            return matchingModuleRoot;
-        }
+        var moduleRoots = ModuleRootManager.getInstance(module).getContentRoots();
+        var projectRoots = ProjectRootManager.getInstance(module.getProject()).getContentRoots();
+        var candidateProjectRoot = Arrays.stream(projectRoots)
+                .filter(contentRoot -> Arrays.stream(moduleRoots).anyMatch(moduleRoot -> {
+                    return VfsUtilCore.isAncestor(contentRoot, moduleRoot, false);
+                }))
+                .min(Comparator.comparingInt(file -> file.getPath().length()));
 
-        // Some project models, e.g. Gradle, create multiple modules for a single gradle project.
-        // For example, Gradle module "main" with content root "<root>/module-one/src/main"
-        // and a top-level module with content root "<root>/module-one".
-        // The context (e.g. "<root>/module-one") may be located outside the root of module (e.g "main").
-        var projectContentRoot = ProjectRootManager.getInstance(module.getProject())
-                .getFileIndex()
-                .getContentRootForFile(context, false);
-        if (projectContentRoot != null) {
-            return projectContentRoot;
+        if (candidateProjectRoot.isPresent()) {
+            return candidateProjectRoot.get();
         }
 
         // fallback to the context itself for unexpected project setups
--- a/plugin-java/src/main/java/appland/execution/AppMapJavaPackageConfig.java
+++ b/plugin-java/src/main/java/appland/execution/AppMapJavaPackageConfig.java
@@ -4,8 +4,7 @@ import appland.config.AppMapConfigFile;
 import appland.files.AppMapFiles;
 import com.intellij.openapi.application.ReadAction;
 import com.intellij.openapi.module.Module;
-import com.intellij.openapi.module.ModuleUtil;
-import com.intellij.openapi.roots.ModuleRootManager;
+import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ProjectRootManager;
 import com.intellij.openapi.util.text.Strings;
 import com.intellij.openapi.vfs.LocalFileSystem;
@@ -21,6 +20,7 @@ import com.intellij.util.PathUtil;
 import com.intellij.util.concurrency.annotations.RequiresReadLock;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.jps.model.java.JavaModuleSourceRootTypes;
 
 import java.io.IOException;
 import java.nio.file.Path;
@@ -52,11 +52,11 @@ final class AppMapJavaPackageConfig {
     }
 
     public static @NotNull Path createAppMapConfig(@NotNull Module module,
-                                                   @NotNull VirtualFile configParentDirectory,
+                                                   @NotNull VirtualFile appMapContentRootDirectory,
                                                    @NotNull Path appMapOutputDirectory) throws IOException {
-        assert (configParentDirectory.isDirectory());
+        assert (appMapContentRootDirectory.isDirectory());
 
-        var configParentPath = configParentDirectory.toNioPath();
+        var configParentPath = appMapContentRootDirectory.toNioPath();
         if (appMapOutputDirectory.isAbsolute() && !appMapOutputDirectory.startsWith(configParentPath)) {
             throw new IllegalStateException(String.format("AppMap output directory is not inside the working directory: %s, %s",
                     configParentPath,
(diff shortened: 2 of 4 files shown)


plugin-java/src/main/java/appland/execution/AppMapJavaPackageConfig.java
```diff
@@ -220,8 +220,13 @@ private static VirtualFile[] collectSourceRootsWithDependencies(@NotNull Module
                 .flatMap(m -> Arrays.stream(ModuleRootManager.getInstance(m).getContentRoots()))
                 .collect(Collectors.toUnmodifiableSet());
 
-        return Arrays.stream(ProjectRootManager.getInstance(module.getProject()).getContentSourceRoots())
-                .filter(sourceRoot -> VfsUtilCore.isUnder(sourceRoot, contentRoots))
+        ProjectRootManager rootManager = ProjectRootManager.getInstance(module.getProject());
```
jansorg, line 223: The new code is filtering all source roots of the project, which means that all source roots of the project are returned. With all source roots, all packages available in the project would be inserted into `appmap.yml`. But only packages contained in `module` (and its dependencies) must be inserted.

The old code is taking care of that by filtering the available roots with variable `contentRoots`, which represents all allowed content root directories. AFAIK a project source directory is always located inside a content root.
Because of the possible nesting of modules, the old code retrieved the content roots of the module and all dependency modules and then filtered the available roots with these roots.

IMHO this should work:
```suggestion
        return ProjectRootManager.getInstance(module.getProject()).getModuleSourceRoots(JavaModuleSourceRootTypes.SOURCES)
                .stream()
                .filter(sourceRoot -> VfsUtilCore.isUnder(sourceRoot, contentRoots))
                .toArray(VirtualFile[]::new);
```

jansorg (changes requested):

plugin-java/src/main/java/appland/execution/AppMapJavaPackageConfig.java
apotterri: > With all source roots, all packages available in the project would be inserted into appmap.yml. But only packages contained in module (and its dependencies) must be inserted.

Can you help me understand why this is, please? What's the relationship between IntelliJ's notion of a project and a module, and how is each related to what's actually on disk (e.g. in a git repo)? If you could relate that to what happens when you open https://github.com/finos/waltz in IntelliJ, that would be a big help.

Thanks!

jansorg: In general, usually a single IntelliJ window represents a project. A project consists of 0..n modules and each module has entities, e.g. top-level content roots (directories containing module data), nested source roots (directories containing source files), resource roots, excluded folders, and more. 
Help pages: 
- https://www.jetbrains.com/help/idea/creating-and-managing-modules.html
- https://www.jetbrains.com/help/idea/content-roots.html

The project model is unrelated to a git repo. A project is often imported from an external build system, e.g. Gradle. A Gradle project and its sub-projects is mapped as 1..n modules. It's possible to add several Gradle or Maven projects into a single IntelliJ project.
Therefore we have to be careful about scoping the search for source roots to avoid collecting packages of unrelated modules (e.g. of a different Gradle project or sub-project).

For waltz, you can see all imported modules at `File>Project Structure>Module`. You can select a module to see the content roots, source roots, etc.
- waltz
- waltz-common
- waltz-data
- ...

With this code:
```java
var fileIndex = rootManager.getFileIndex();
        VirtualFile[] contentSourceRoots = rootManager.getContentSourceRoots();
        return Arrays.stream(contentSourceRoots)
                .filter(sourceRoot -> {
                    return fileIndex.isUnderSourceRootOfType(sourceRoot, JavaModuleSourceRootTypes.SOURCES);
                })
                .toArray(VirtualFile[]::new);
```

`rootManager` is a `ProjectRootManager`, which is looking at all modules at once and not just `module`.
A Java package is always located under a source root. If you collect all source roots, then you'll get all of the project's packages.

apotterri: Ok, thanks, I think I understand this explanation.

I'm not sure either of our solutions gets us the result we want, though.

I created this project by importing spring-petclinic as a module in my waltz project:
image

If I run, e.g. the `test` goal in `waltz-jobs`, we want to create `appmap.yml` in the root directory of the local git repo. It should contain all the packages from the all the modules, i.e. the same content as when the `test` goal of the `waltz` module is executed.

Does that make sense?

jansorg: I think we should only use the project-module (i.e. modules, content roots, source roots, etc.) to create appmap.yml and not data like a `.git` directory, which isn't used by the IDE itself to create such model. There may be all kinds of projects without git and the project model contains all we need, I believe.

For waltz, I think I know what's going on.
For Gradle-based multi-module projects (e.g. junit5), the working directory of the Gradle run configuration is the top-level directory. Because it's the top-level dir, the file `appmap.yml` is created in the correct directory.

For waltz, which is Maven-based, the working directory (or, rather, the context module of the run configuration) is the sub-project, e.g. `waltz-jobs`. Therefore `appmap.yml` is created in the sub-directory.
For a checkout of `waltz` at `/sources/waltz`, you can see that:
- top-level module `waltz` has content root `/sources/waltz`
- sub-module `waltz-jobs` has content root `/sources/waltz/waltz-jobs`.

We need changes to the code to make this work for waltz and other setups.

IMHO this should work, using waltz as example
- Find the run configuration context, in our case it's `waltz-jobs` with content root `/sources/waltz/waltz-jobs`
- We would like to create `appmap.yml` in the waltz top-level dir `/sources/waltz`, which belongs to module `waltz`. 
A generic solution would be to iterate the content roots of the project's modules and take the root, which is parent of the start location (e.g. `/sources/waltz/waltz-jobs`) and has no other content root as parent.
In our example, this is `/sources/waltz` of module `waltz`.
- Then the Java packages need to be searched in the module `waltz` plus all modules with a content root below `/sources/waltz` (`waltz-common`, `waltz-data`, `waltz-jobs`, etc.). For correctness, we probably need to look at the dependencies of these modules, too.

I think it would be a small change, but because we had to patch this code a few times already, let's think about it before our next attempt :)

kgilpin: Thanks, and I agree about using project information rather than git.

apotterri: Yeah, I only referenced the git repo because I didn't know how else to say that `appmap.yml` needs to get created in the root of the project.

plugin-java/src/main/java/appland/execution/AppMapJavaConfigUtil.java
```diff
@@ -50,24 +33,18 @@ private AppMapJavaConfigUtil() {
...
-        // For example, Gradle module "main" with content root "<root>/module-one/src/main"
-        // and a top-level module with content root "<root>/module-one".
-        // The context (e.g. "<root>/module-one") may be located outside the root of module (e.g "main").
-        var projectContentRoot = ProjectRootManager.getInstance(module.getProject())
-                .getFileIndex()
-                .getContentRootForFile(context, false);
-        if (projectContentRoot != null) {
-            return projectContentRoot;
+        // Any of the module's roots will do
+        var moduleRoot = ModuleRootManager.getInstance(module).getContentRoots()[0];
```
jansorg, line 37: I don't think there's a guarantee that a module contains at least one content root.

plugin-java/src/main/java/appland/execution/AppMapJavaConfigUtil.java
```diff
@@ -50,24 +33,18 @@ private AppMapJavaConfigUtil() {
...
+        var projectRoots = ProjectRootManager.getInstance(module.getProject()).getContentRoots();
+        var candidateProjectRoots = Arrays.stream(projectRoots)
+                .filter(sourceRoot -> VfsUtilCore.isAncestor(sourceRoot, moduleRoot, true))
```
jansorg, line 40: Passing `true` here means that `moduleRoot` must be flow `sourceRoot` and can't be the same. I don't think we have to enforce this and could allow module roots, which equal `sourceRoot` here.
(Technically, `sourceRoot` should be named `contentRoot`)

plugin-java/src/main/java/appland/execution/AppMapJavaConfigUtil.java
```diff
@@ -50,24 +33,18 @@ private AppMapJavaConfigUtil() {
...
-        // and a top-level module with content root "<root>/module-one".
-        // The context (e.g. "<root>/module-one") may be located outside the root of module (e.g "main").
-        var projectContentRoot = ProjectRootManager.getInstance(module.getProject())
-                .getFileIndex()
-                .getContentRootForFile(context, false);
-        if (projectContentRoot != null) {
-            return projectContentRoot;
+        // Any of the module's roots will do
+        var moduleRoot = ModuleRootManager.getInstance(module).getContentRoots()[0];
+        var projectRoots = ProjectRootManager.getInstance(module.getProject()).getContentRoots();
```
jansorg, line 38: IMHO it would be better to look at all module content roots (even though most will contain 1 root only).
There's a `min` method on streams to help choosing the shortest path.

Perhaps something like the following?
```java
        var moduleRoots = ModuleRootManager.getInstance(module).getContentRoots();
        var projectRoots = ProjectRootManager.getInstance(module.getProject()).getContentRoots();
        var candidateProjectRoot = Arrays.stream(projectRoots)
                .filter(contentRoot -> Arrays.stream(moduleRoots).anyMatch(moduleRoot -> {
                    return VfsUtilCore.isAncestor(contentRoot, moduleRoot, false);
                }))
                .min(Comparator.comparingInt(file -> file.getPath().length()));

        if (candidateProjectRoot.isPresent()) {
            // The candidate roots are sorted, so  the top-level one will be
            // first (because it will be shortest).
            return candidateProjectRoot.get();
        }
```

But I'm currently not sure, what would happen if we returned a project content root, which does not contain the context file. Have you seen something like this or what do you think about this edge-case?

plugin-java/src/main/java/appland/execution/AppMapJavaPackageConfig.java
```diff
@@ -220,7 +224,9 @@ private static VirtualFile[] collectSourceRootsWithDependencies(@NotNull Module
                 .flatMap(m -> Arrays.stream(ModuleRootManager.getInstance(m).getContentRoots()))
                 .collect(Collectors.toUnmodifiableSet());
 
-        return Arrays.stream(ProjectRootManager.getInstance(module.getProject()).getContentSourceRoots())
+        return ProjectRootManager.getInstance(module.getProject())
```
jansorg, line 227: The changes for this PR are small, but complex.
I like the change to find the best matching AppMap content root. 

After that directory is found, this method `collectSourceRootsWithDependencies` is called with the generic module again. I think that's not sufficient. 
For example, if you executed `tests` inside `waltz-jobs` of the  `waltz` project, the selected "best AppMap content root" would be `/source/waltz`, but here the code is limiting the packages search for module `waltz-jobs` and its dependencies. This would exclude code of the other modules, for example `waltz-common`.

Instead, using the selected "Best AppMap content root" to filter the available source roots of the project should (hopefully) provide the suitable source roots. 

Because I'm not good at explaining this, I've made https://github.com/getappmap/appmap-intellij-plugin/pull/702 to show some code. Please feel free to use it or to make your own changes where necessary.

(There's a possible edge-case for modules, which have several source roots and which have source roots under "Best AppMap content root" and also source roots outside the "Best AppMap content root". But I'd like to get your opinion on the notes above before we're discussing rather weird edge-cases).

jansorg (changes requested):
Thanks!
I've added a few notes and suggestions in the review. Please let me know if you need more details.

apotterri:
Thanks for reviewing these changes. I pulled in the suggestions you made.

As far as edge cases go, I feel like we've already spent enough time on this, maybe more than it really merited. I think we've got the majority of the cases covered. If we run into a case where we can't derive the correct location for `appmap.yml` from the project, we'll fall back to creating it in the current working directory. This seems to me like it's good enough for now.

jansorg (approved):
Okay, let's go with this.
We can fix edge-cased when users report them.

Issue #700, opened 2024-05-08
HTTP proxy configuration is propagated to the RPC service


dustinbyrne:
`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`

PR #703, opened 2024-05-10
Pass IDE proxy settings to the AppMap CLI tools

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/700

With this PR, the IDE's HTTP proxy settings are passed as environment variables to all AppMap CLI processes launched by the plugin. This include the JSON-RPC server, scanner and indexer.

The following variables are added to the environment of AppMap processes, but only if HTTP proxy settings of the IDE are enabled:
- `http_proxy`
- `https_proxy` 
- `no_proxy` 

If the IDE configured username and password for proxy authentication, then both are included in the proxy URL values, e.g. `http://myUser:secure_password@my.proxy:1234`.

I'm not aware of a way to listen to changes of the IDE's proxy settings.

For unknown reasons the download tests started to fail, https://github.com/getappmap/appmap-intellij-plugin/pull/703/commits/27acefa43e65fbe6a3ba025d852f376489172675 is fixing this. I don't know how this could be related to AppMap proxy settings, I think it's either slow(er) runners or a change in another component used by CI.

### DIFF
--- a/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
+++ b/plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
@@ -20,12 +20,14 @@ import com.intellij.openapi.project.DumbService;
 import com.intellij.openapi.project.ProjectManager;
 import com.intellij.openapi.util.Key;
 import com.intellij.openapi.util.SystemInfo;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.openapi.vfs.ex.temp.TempFileSystem;
 import com.intellij.util.concurrency.AppExecutorUtil;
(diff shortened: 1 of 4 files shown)


plugin-core/src/main/java/appland/cli/DefaultCommandLineService.java
```diff
@@ -572,7 +577,57 @@ private static GeneralCommandLine applyServiceEnvironment(@NotNull GeneralComman
...
+            var noProxy = settings.PROXY_EXCEPTIONS;
+            return Map.of(
+                    // lower-cased names
+                    "http_proxy", proxyURL,
+                    "https_proxy", proxyURL,
+                    "no_proxy", noProxy,
+                    // upper-cased names
+                    "HTTP_PROXY", proxyURL,
+                    "HTTPS_PROXY", proxyURL,
+                    "NO_PROXY", noProxy
```
kgilpin, line 625: It's fine to do just upper or lower case, we will check for both in the RPC service.

dustinbyrne (approved):

PR #713, opened 2024-05-15
fix: Drop mode buttons

dustinbyrne:


kgilpin:
Thanks, can you check that we are not using @generate or any other @ commands on the listing / description page?

PR #726, opened 2024-06-11
Show warning for users of 2024.1 with proxy servers

jansorg:
This PR displays a message about broken proxy support for users
- using 2024.1.x
- with an enabled HTTP proxy server
- which did not click "Don't show again" when the same message was displayed earlier

"Don't show again" deactivates the message for all projects.

### DIFF
--- a/plugin-core/src/main/java/appland/notifications/AppMapNotifications.java
+++ b/plugin-core/src/main/java/appland/notifications/AppMapNotifications.java
@@ -3,15 +3,20 @@ package appland.notifications;
 import appland.AppMapBundle;
 import appland.AppMapPlugin;
 import appland.actions.StopAppMapRecordingAction;
+import appland.settings.AppMapApplicationSettingsService;
 import appland.startup.FirstAppMapLaunchStartupActivity;
 import appland.webviews.navie.NavieEditorProvider;
 import com.intellij.ide.BrowserUtil;
 import com.intellij.notification.*;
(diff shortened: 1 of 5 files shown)


plugin-core/src/main/java/appland/notifications/AppMapNotifications.java
```diff
@@ -160,6 +165,44 @@ public static void showNavieUnavailableNotification(@NotNull Project project) {
...
+                    new DialogWrapper.DoNotAskOption.Adapter() {
+                        @Override
+                        public void rememberChoice(boolean isSelected, int exitCode) {
+                            AppMapApplicationSettingsService.getInstance().setShowBrokenProxyWarning(!isSelected);
+                        }
+                    }
+            );
+
+            if (action == 1) {
+                BrowserUtil.open("https://appmap.io/status");
```
dustinbyrne, line 190: I think we should drop this. This page makes no mention of the issue described in the dialog. Let's just offer a 'close' button, the dialog content is sufficient.

dustinbyrne (changes requested):

plugin-core/src/main/java/appland/notifications/AppMapNotifications.java
jansorg: @dustinbyrne I've rebased and removed the button to open the status page.
image

dustinbyrne (approved):

Issue #557, opened 2024-02-05
Run tests on GitHub Actions against earliest and latest SDK


jansorg:
As discussed at https://github.com/getappmap/appmap-intellij-plugin/pull/553, it would be helpful to not just run tests against the earliest supported SDK.

PR #735, opened 2024-07-02
Run tests against the latest supported major version

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/557
CI: Only the AppMap analysis failed and I'm unable to fix it because a re-run doesn't do that.

This PR adds an additional matrix build configuration to test against the latest supported IDE major version.
To avoid unnecessarily long job runtime, we're only running tests against the latest version on Linux, for now.

The verification of the plugin structure and binary release was split off into another job to simplify the structure of the workflow.

Tests, which failed on 2024.1 but not on 2023.1, are also fixed in this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
               run: ./gradlew verifyPlugin
 
             - name: Upload Test Reports
-              uses: actions/upload-artifact@v3
+              uses: actions/upload-artifact@v4
               if: always()
               with:
                   name: test-reports-${{ matrix.os }}
@@ -63,7 +63,7 @@ jobs:
                       **/build/reports/
 
             - name: Save AppMaps
-              uses: actions/cache/save@v3
+              uses: actions/cache/save@v4
               if: runner.os == 'Linux'
               with:
                   path: ./tmp/appmap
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -39,43 +45,49 @@ jobs:
...
+
+            - name: Setup Java
+              uses: actions/setup-java@v4
+              with:
+                  distribution: temurin
+                  java-version: ${{ env.JAVA_VERSION }}
+
+            - name: Verify Plugin Structure
+              shell: bash
+              run: ./gradlew verifyPlugin
```
dustinbyrne, line 85: Only somewhat related, but this is failing on main
https://github.com/getappmap/appmap-intellij-plugin/actions/runs/9914172346/job/27392737457#step:8:1

jansorg: https://github.com/getappmap/appmap-intellij-plugin/pull/742 should fix the CI failure

jansorg: CI of this PR is failing atm because the current CI expired (I think). I'll rebase when #743 is merged.

dustinbyrne (approved):

PR #775, opened 2024-08-12
Avoid exception about slow operation when loading OpenAI key

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/773

In 2024.2+ the method to load a key from secure password storage is a slow operation and causes an exception.
This PR applies the changes to load the key in a background task to avoid the exception.
The key is cached in memory after it's been retrieved from secure storage to avoid the expensive call at runtime.

### DIFF
--- a/plugin-core/src/main/java/appland/settings/AppMapSecureApplicationSettingsService.java
+++ b/plugin-core/src/main/java/appland/settings/AppMapSecureApplicationSettingsService.java
@@ -1,10 +1,14 @@
 package appland.settings;
 
+import appland.AppMapBundle;
 import com.intellij.credentialStore.CredentialAttributes;
 import com.intellij.credentialStore.CredentialAttributesKt;
 import com.intellij.ide.passwordSafe.PasswordSafe;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.components.Service;
+import com.intellij.openapi.diagnostic.Logger;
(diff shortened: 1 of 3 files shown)


plugin-core/src/main/java/appland/settings/AppMapSecureApplicationSettingsService.java
```diff
@@ -16,16 +20,57 @@ public final class AppMapSecureApplicationSettingsService implements AppMapSecur
...
     }
 
     @Override
     public void setOpenAIKey(@Nullable String key) {
         var oldValue = getOpenAIKey();
         try {
             PasswordSafe.getInstance().setPassword(createOpenAIKey(), key);
+
+            cachedOpenAIKey = key;
+            isCached = true;
```
dustinbyrne, line 73: nit: `cachedOpenAIKey` is nullable, indicating whether or not a key has been cached. Maintaining the `isCached` flag isn't necessary.

dustinbyrne (approved):

plugin-core/src/main/java/appland/settings/AppMapSecureApplicationSettingsService.java
jansorg: I also wanted to cache null values, i.e. the state when there's no stored OpenAI key. That's why I added the `isCached` flag.

PR #786, opened 2024-09-17
Implement pinned files for the Navie editor

jansorg:


apotterri (approved):
Thanks for adding the tests.

PR #796, opened 2024-10-04
Update to use new file pinning UI from @appland/components

### LINKED ISSUE #795: The plugin shows the new UI for pinned files

apotterri:
Fixes #795 .


plugin-core/src/main/kotlin/appland/settings/AppMapProjectSettingsPanel.kt
```diff
@@ -68,6 +72,9 @@ class AppMapProjectSettingsPanel {
                 label(AppMapBundle.get("projectSettings.maxPinnedFileSize.unit"))
                 rowComment(AppMapBundle.get("projectSettings.maxPinnedFileSize.comment"))
             }
+            row {
+                useAnimation = checkBox("Use animation").component
```
jansorg, line 76: This should use i18n via AppMapBundle.get(…)

plugin-core/src/main/java/appland/webviews/navie/NavieEditor.java
```diff
@@ -242,6 +255,22 @@ public void run(@NotNull ProgressIndicator indicator) {
         }.queue();
     }
 
+    private void handleChooseFilesToPin() {
+        DataContext dataContext = dataId -> {
```
jansorg, line 259: Instead of the providing a custom data context and emulating how actions are called, I think it would be better to extract a static method in ChooseAndAddNavieContextFilesAction, because its signature would show the needed input data. Providing input via the generic data context is prone to errors.

if a data context was used, it should be created with our DataContexts because of compatibility with the recent SDKs.

jansorg (changes requested):
Thanks! I added two notes.
I haven‘t run this, I could do that on Monday if needed.

apotterri:
It would be very helpful if you could give these changes a test run, whenever you have a chance. More testing is better. :)

jansorg:
@apotterri I've tested this on Linux and it's working as expected.
I noticed that `ChooseAndAddNavieContextFilesAction` is not checking the file size of the pinned files, but we probably should fix this in another PR because it's unrelated to this PR.

appland-navie/index.html
```diff
@@ -1,58 +1,66 @@
...
+
+<head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>AppMap AI: Explain</title>
+    <link rel="stylesheet" href="./dist/main.css" />
+    <style>
+        body {
+            font-family: system-ui, sans-serif;
+            font-size: 80%;
```
jansorg, line 12: What's the reason for this change?

jansorg (approved):
Thanks, the latest changes seem fine to me!
I wasn't sure why you changed the global font-size for Navie. Was this related to my previous review?

appland-navie/index.html
apotterri: No, this wasn't related to your review. Dan noticed that the text is much bigger than what's rendered in VS Code. This makes them much more similar in size.

Do you see a problem with this change?

jansorg: With this change, on my setup, Navie looked fine to me.
The IDE does have its own scaling factor for UI, but AFAIK it's not applied to the webviews. It's messy, unfortunately. 
There's also `zoom: ...` as a possible alternative  https://developer.mozilla.org/en-US/docs/Web/CSS/zoom.
I'm fine to merge this as it is. I'll keep an eye on the webview scaling on different setups.

apotterri:
[Just noticesd that I hadn't submitted this comment. FYI]

`ChooseAndAddNavieContextFilesAction` calls `NavieEditor.addPinnedFiles` which validates the file size. (I confirmed that using it to try to add a file > 20KB fails as expected.)

appland-navie/index.html
apotterri: I considered `zoom`. It resizes everything in the viewport, rather than just the text, and I don't think we want to do that. Diagrams look ok to me.

jansorg:
Thanks, I noticed this now, too :) I'll push a small PR with clearer documentation on the method.

Issue #816, opened 2024-12-09
Allow Navie to use a user's LLM


jansorg:
https://github.com/getappmap/vscode-appland/pull/983/files#diff-3606d80ad0d196e57c8e0c0bab40cdfa917775d2b2f4074a5a8ecdb04f778eb2

PR #820, opened 2024-12-18
feat: integrate Navie with GitHub Copilot

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/816


plugin-copilot/src/main/java/appland/copilotChat/NavieCopilotChatRequestHandler.java
```diff
@@ -0,0 +1,266 @@
...
+        if (!requestPath.startsWith(BASE_PATH + "/") && !requestPath.equals(BASE_PATH)) {
+            return false;
+        }
+
+        var requestName = requestPath.substring(BASE_PATH.length());
+
+        if ("/chat/completions".equals(requestName)) {
+            // verify the authorization header, which must contain our random ide session id
+            var expectedHeaderValue = "Bearer " + GitHubCopilotService.RandomIdeSessionId;
+            if (!StringUtil.equals(fullHttpRequest.headers().get(HttpHeaderNames.AUTHORIZATION), expectedHeaderValue)) {
```
dividedmind, line 61: Note while perhaps it doesn't matter terribly here, using standard string comparison to compare cryptographic information is a bad idea; because regular string comparison short circuits on first mismatch, an attacker can measure the time to rejection and learn which character was wrong. This allows bruteforcing character-by-character basically defeating any security.

It's a good idea to always use constant-time comparison when dealing with thing like this. [StackOverflow seems to suggest](https://stackoverflow.com/a/37737277) that `java.security.MessageDigest.isEqual` is what you need.

plugin-copilot/src/main/java/appland/copilotChat/NavieCopilotChatRequestHandler.java
```diff
@@ -0,0 +1,266 @@
...
+                var content = Unpooled.copiedBuffer("data: [DONE]\n\n", StandardCharsets.UTF_8);
+                channel.writeAndFlush(new DefaultHttpContent(content));
+
+                var future = channel.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT);
+                future.addListener(ChannelFutureListener.CLOSE);
+            }
+        };
+
+        var response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
+        response.headers().set(HttpHeaderNames.CONTENT_TYPE, "application/json");
```
dividedmind, line 210: I believe it should be `application/event-stream` when streaming.

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotChatSession.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private static void processSSEChunk(@NotNull CopilotChatResponseListener responseListener,
+                                        @NotNull String sseData) {
+        var chunk = GitHubCopilotService.gson.fromJson(sseData, CopilotChatCompletionsStreamChunk.class);
+
+        var id = chunk.id();
+        var modelName = chunk.model();
+        var created = chunk.created();
+        for (var choice : chunk.choices()) {
```
dividedmind, line 107: This isn't right. You only want the first choice. (Although In general with default parameters there will be only one.)

dividedmind:
I know it's a draft, but I gave it a once-over and I noticed you seem to be directly using GitHub's undocumented APIs instead of calling out to the language server as we discussed. I worry a bit that this approach might be more fragile in case they change something, or frown upon this usage and block us out. Also, their API seems to be OpenAI-compatible — since that's the case, is there a reason to parse and rebuild all the responses instead of just passing along the requests and replies verbatim, filling in any necessary headers?

jansorg:
Thanks for your feedback!

> I know it's a draft, but I gave it a once-over and I noticed you seem to be directly using GitHub's undocumented APIs instead of calling out to the language server as we discussed. 

As far as I understand, this wasn't the conclusion of our discussions. Or, rather, there was no clear conclusion IMHO.

The first attempt was to use the Java interfaces. But this did not work out, as there's no open API and many undocumented classes to handle the chat. 

Using the LSP (via non-standard LSP commands) may be an alternative, as we discussed.
AFAIK there's no documentation of the LSP commands, chat is not part of the current LSP spec and using undocumented LSP seemed more difficult to me.
IMHO it's subject to change in the same way as the HTTP API and would provide a similar (sometimes unstable) interface.
I couldn't find code using LSP commands in the official Copilot plugin, but it's possible that I missed it.
Chat seems to be not yet supported, https://github.com/orgs/community/discussions/50939#discussioncomment-11439257

The HTTP API seems to work okay for the Neovim Chat plugin https://github.com/CopilotC-Nvim/CopilotChat.nvim/

> I worry a bit that this approach might be more fragile in case they change something, or frown upon this usage and block us out. Also, their API seems to be OpenAI-compatible — since that's the case, is there a reason to parse and rebuild all the responses instead of just passing along the requests and replies verbatim, filling in any necessary headers?

You're right that the responses are very similar.
It's perhaps possible to just forward the JSON chunks, but to learn and understand the Copilot API I preferred to parse the data and convert it.
At the end, when the implementation is feature complete, I'm fine to decide which approach works best. Forwarding JSON would require that it's exactly what Navie's OpenAPI needs, parsing and converting is more flexible (and more prone to errors).

There are minor differences in roles, for example. https://github.com/getappmap/appmap-intellij-plugin/blob/a513cf400833fe47c8e0b875baee281d2a72e1e1/plugin-copilot/src/main/java/appland/copilotChat/openAI/OpenAIChatRoles.java#L13
And, I think, Copilot is not providing usage data as expected by Navie's use of OpenAPI, https://github.com/getappmap/appmap-intellij-plugin/blob/a513cf400833fe47c8e0b875baee281d2a72e1e1/plugin-copilot/src/main/java/appland/copilotChat/openAI/OpenAICompletionsStaticResponse.java#L15

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotChatSession.java
jansorg: Thanks, good point. I'll fix that when I'm back.

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotChatSession.java
```diff
@@ -0,0 +1,125 @@
...
+                .tuner(connection -> {
+                    applyHeaders(connection, baseHeaders);
+                    connection.setRequestProperty("openai-intent", "conversation-panel");
+                    connection.setRequestProperty("openai-organization", "github-copilot");
+                    connection.setRequestProperty("x-request-id", requestId());
+                })
+                .connect(httpRequest -> {
+                    httpRequest.write(GitHubCopilotService.gson.toJson(copilotRequest));
+                    readServerEvents(httpRequest.getReader(), responseListener);
+                    return null;
```
dustinbyrne, line 68: It's possible that the remote endpoint returns an error. In this case, the error should propagate to the client as a response containing a JSON object.

Property | Type | Description
-- | -- | --
error | Object | Encapsulates the error details.
error.message | String | A detailed message describing the error.
error.type | String | The type or category of the error.
error.param | String or null | (Optional) Parameter associated with the error.
error.code | String or null | (Optional) Specific error code for machine readability.

Here's the implementation in VSCode where we coerce Error types into OpenAI-compliant JSON error objects:
https://github.com/getappmap/vscode-appland/blob/bfd83ad8c848d31257ab004688eb847feecbcf32/src/services/chatCompletion.ts#L359

If the request has exceeded the token limit, it's quite important that the error message match the implementation above. The OpenAI error reporting a token overflow only states the number of used and max tokens in the error message, so this information is parsed out elsewhere. I don't have experience with this particular endpoint, so it's possible you don't actually need to manipulate the message in practice. It's also possible that the endpoint is also returning the desired response type, it just needs to be forwarded along.

You can replicate this easily by pasting in some lengthy content. Currently, an internal IDE error is raised.

dustinbyrne:
This style can now be removed, it hides the Copilot option from the LLM configuration menu.
https://github.com/getappmap/appmap-intellij-plugin/blob/58ff3623a0f9bf3b72c5bb5f016bf4bfcbb281d0/appland-navie/index.html#L59-L61

image

I've noticed that when Copilot is enabled, there's (seemingly?) no way to disable it. Can the LLM configuration options work with Copilot?

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotChatSession.java
jansorg: I've changed this to send all choices in a single chunk.

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotChatSession.java
jansorg: I've implemented error response handling, https://github.com/getappmap/appmap-intellij-plugin/blob/bb11a0e1d6b456d082e7264f1beec79811ed82f0/plugin-copilot/src/main/java/appland/copilotChat/NavieCopilotChatRequestHandler.java#L289

jansorg:
> This style can now be removed, it hides the Copilot option from the LLM configuration menu.

@dustinbyrne  I've removed the style and also made a change to make the LLM chooser work if Copilot is enabled. It currently requires a restart of the AppMap RPC service.

plugin-core/src/main/java/appland/webviews/navie/NavieEditor.java
```diff
@@ -329,11 +329,15 @@ private void handleSelectLlmOption(@NotNull String choice) {
         switch (choice) {
             // Clear LLM environment settings and remove OpenAPI key
             case "default": {
```
dustinbyrne, line 331: This works for switching between the default hosted provider and Copilot, but I'm not able to select "use your own OpenAI API key". After reloading the project, it defaults back to Copilot. I haven't tested it, but I imagine "use your own model" would similarly fall back to Copilot.

jansorg: Thanks! That's a bug.
I've fixed it and it's part of the latest preview build.

plugin-copilot/src/main/java/appland/copilotChat/CopilotAppMapEnvProvider.java
```diff
@@ -26,13 +27,23 @@ public Map<String, String> getEnvironment() {
...
-    public static boolean isGitHubCopilotEnabled() {
-        if (!AppMapApplicationSettingsService.getInstance().isEnableCopilotIntegration()) {
-            return false;
+    /**
+     * @return {@code true} if the user has set custom model settings for Navie for an OpenAI or Azure OpenAI API key.
+     * Existing settings for a custom key override the Copilot integration.
+     */
+    public static boolean hasCustomModelSettings() {
+        var environment = AppMapApplicationSettingsService.getInstance().getCliEnvironment();
+        return environment.containsKey(AppLandJsonRpcService.OPENAI_API_KEY)
```
dustinbyrne, line 40: This works when `OPENAI_API_KEY` is added to the environment variables setting, but not when the key is set via the 'OpenAI key' setting (or selecting "use your own OpenAI API key" from the LLM configuration menu).

I'd also add `OPENAI_BASE_URL` to this list of variables that indicate custom model settings.

jansorg: Thanks! I've fixed both issues.

plugin-copilot/src/main/java/appland/copilotChat/CopilotAppMapEnvProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+        return Map.of(
+                "OPENAI_BASE_URL", NavieCopilotChatRequestHandler.getBaseUrl(),
+                "APPMAP_NAVIE_COMPLETION_BACKEND", "openai",
+                "OPENAI_API_KEY", GitHubCopilotService.RandomIdeSessionId
+        );
+    }
+
+    public static boolean isGitHubCopilotIntegrationEnabled() {
+        return !hasCustomModelSettings() && !isGitHubCopilotDisabled();
```
dustinbyrne, line 32: Is it possible that this returns `true` yet Copilot has no models available via the API? If so, we should consider this case.

jansorg: Hm, if Copilot happens to return zero models, this may be the case.
But I don't think that Copilot's API could work without any model.

A model would be requested from Copilot when Navie is being used, i.e. possibly much later than this method is being called.

dividedmind:
I observed that when Copilot is installed but not logged in, Navie still tries to use Copilot and just hangs there.

Specifically, here's what I did:
- logged into copilot, verified that Navie works,
- logged out — weirdly, Navie still works,
- restarted IntelliJ — Copilot is logged out, Navie is still in Copilot mode, but any question just hangs there.

It's possible that I used the backend configurator to switch to hosted model and back at some point, but I don't know if it was significant.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -0,0 +1,144 @@
...
+            LOG.debug("Failed to read GitHub Copilot config: " + configPath, e);
+            return null;
+        }
+    }
+
+    /**
+     * @return The {@link Path} to the GitHub Copilot plugin's configuration file, or {@code null} if it could not be found.
+     */
+    private static @Null Path findGitHubCopilotConfig() {
+        var basePath = Path.of(SystemProperties.getUserHome(), ".config", "github-copilot");
```
dividedmind, line 115: Will this work on Windows?

jansorg: It should, but I'll test again on Windows now.

jansorg: @dividedmind Thanks for raising this, it did not work on Windows 11. I'll fix this now.

PR #841, opened 2025-01-22
Support to choose the copilot model for Navie

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/824

This PR adds a new action to choose the Copilot model for Navie.

This changes the existing "loadModels" to only return models suitable for chat, i.e. only models with `type == "chat"`.

Menu item to invoke the action:
image

Popup to choose the model. The model currently selected for Navie is pre-selected when the popup is shown.
image

### DIFF
--- a/plugin-copilot/src/main/java/appland/copilotChat/ChooseCopilotModelAction.java
+++ b/plugin-copilot/src/main/java/appland/copilotChat/ChooseCopilotModelAction.java
@@ -1,0 +1,157 @@
+package appland.copilotChat;
+
+import appland.AppMapBundle;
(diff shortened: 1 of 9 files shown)


kgilpin:
You’ve probably noticed, but some of the model names are duplicated/redundant.

jansorg:
> You’ve probably noticed, but some of the model names are duplicated/redundant.

Yes, I noticed. It's caused by models with the same name and version, but different ID. I think GitHub is keeping older models (e.g. preview models) around even after they moved into production mode. 
I'll remove the duplicates in the final version.

dustinbyrne (approved):

Issue #844, opened 2025-01-23
RPC service is automatically restarted on configuration change.


dustinbyrne:
#### Overview

The RPC service must be capable of restarting when environment variables change, while maintaining a stable port. The frontend must be notified when the service is restarting and when it has restarted, so it can perform specific actions. With this change active, the user should no longer need to close their project after configuration changes to Navie are made.

#### Requirements

1. **Environment Change Detection**:
   - Monitor environment variables for changes (e.g., user switches backend (Copilot, OpenAI API key, hosted), changes Copilot model, or modifies environment in settings). 
   - Trigger a restart of the RPC service when a change is detected.

2. **Stable Port Maintenance**:
   - Ensure the RPC service restarts using the last known port.

3. **Frontend Notification**:
   - Notify the frontend when the RPC service is restarting.
   - Notify the frontend when the RPC service has restarted.

#### Behavior

1. **Monitoring Environment Variables**:
   - Continuously monitor the environment for changes.
   - Use a debounce mechanism, if needed, to avoid multiple rapid restarts.

2. **Restarting the RPC Service**:
   - When a change is detected, initiate the restart process.
   - Notify the frontend that the service is restarting.
   - Stop the current RPC service instance.
   - Start a new instance of the RPC service using the last known port.
   - Notify the frontend that the service has restarted.

3. **Frontend Interaction**:
   - When the service restart first initiates, notify the frontend. The frontend should respond by calling `onNavieRestarting()` on the Vue app. [Example](https://github.com/getappmap/vscode-appland/blob/076746891e254ca46a592658641918253d0c81ed/web/src/chatSearchView.js#L88-L90)
   - When the service comes back up (indicated by it logging the port its listening on), the frontend is notified. The frontend should respond by calling `loadNavieConfig()` on the Vue app. [Example](https://github.com/getappmap/vscode-appland/blob/develop/web/src/chatSearchView.js#L91-L93)

PR #847, opened 2025-01-29
Restart JSON-RPC server and update Navie without reopening the project

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/844


dividedmind (approved):
I've noticed that "choose copilot model" dialog still tells the user to restart; this should probably be removed (and there might be other places like this).

Other than that — looks good, thanks!

jansorg:
Thanks for the review!

> I've noticed that "choose copilot model" dialog still tells the user to restart; this should probably be removed (and there might be other places like this).

This has to be fixed in the webview UI, I think.
I created https://github.com/getappmap/appmap-js/issues/2251 to track this.

dividedmind:
> Thanks for the review!
> 
> > I've noticed that "choose copilot model" dialog still tells the user to restart; this should probably be removed (and there might be other places like this).
> 
> This has to be fixed in the webview UI, I think. I created [getappmap/appmap-js#2251](https://github.com/getappmap/appmap-js/issues/2251) to track this.

I don't think so — the message is here: https://github.com/getappmap/appmap-intellij-plugin/blob/3fc55626e999be0f111526d0a40c0957127518d8/plugin-core/src/main/resources/messages/appland.properties#L76

PR #854, opened 2025-02-05
fix: Forward inbound request headers to Copilot

dustinbyrne:
This change improves the handling of HTTP headers when making requests to GitHub Copilot by:

1. Forwarding all inbound request headers from the client to Copilot
2. Removing the hardcoded user-agent header since it should be set by the requesting entity

The implementation collects all headers from the incoming request and applies them to the outbound Copilot request, while maintaining existing required headers for Copilot authentication and session management.

### DIFF
--- a/plugin-copilot/src/main/java/appland/copilotChat/NavieCopilotChatRequestHandler.java
+++ b/plugin-copilot/src/main/java/appland/copilotChat/NavieCopilotChatRequestHandler.java
@@ -26,6 +26,7 @@ import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * Adds a new request handler to the IDE's built-in HTTP server.
@@ -103,6 +104,7 @@ public class NavieCopilotChatRequestHandler extends HttpRequestHandler {
             if (openAIRequest.stream()) {
                 sendStreamingChatResponse(chatSession,
                         copilotModel,
+                        fullHttpRequest,
                         openAIRequest,
                         channelHandlerContext);
             } else {
@@ -120,6 +122,7 @@ public class NavieCopilotChatRequestHandler extends HttpRequestHandler {
 
     private void sendStreamingChatResponse(@NotNull CopilotChatSession chatSession,
                                            @NotNull CopilotModelDefinition copilotModel,
+                                           @NotNull FullHttpRequest fullHttpRequest,
                                            @NotNull OpenAIChatCompletionsRequest openAIRequest,
                                            @NotNull ChannelHandlerContext context) throws IOException {
         var channel = context.channel();
@@ -151,6 +154,7 @@ public class NavieCopilotChatRequestHandler extends HttpRequestHandler {
         chatSession.ask(converter,
                 copilotModel,
                 asCopilotChatMessages(openAIRequest.messages()),
+                collectProxiedRequestHeaders(fullHttpRequest),
                 openAIRequest.temperature(),
                 openAIRequest.topP(),
                 openAIRequest.n());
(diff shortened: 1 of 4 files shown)


plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -72,7 +72,6 @@ public boolean isCopilotAuthenticated() {
                 "copilot-language-server-version", GitHubCopilot.LANGUAGE_SERVER_VERSION,
                 "editor-plugin-version", GitHubCopilot.GITHUB_COPILOT_PLUGIN_VERSION,
                 "editor-version", getCopilotEditorVersion(),
```
kgilpin, line 74: Add AppMap plugin version?

dustinbyrne:
I'll prefix existing headers with `x-appmap` and add a plugin version header.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
kgilpin: We can prefix these with `X-AppMap`

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -72,7 +72,6 @@ public boolean isCopilotAuthenticated() {
...
-                "user-agent", GitHubCopilot.USER_AGENT,
```
jansorg, line 75: This user agent header is sent by the GitHub Copilot plugin, I wanted to use it in the same way to maximize compatibility.
It's unknown if/how the backend is using it, but of the requests are still working, I'm okay to remove it.

jansorg (approved):
Seems fine to me. I added a note.
I'll only be able to test this later today, feel free to merge now.

kgilpin:
One thing we may want to add here is something that's unique to the user, like maybe a digest of the HOME directory?

jansorg:
@dustinbyrne I tested this PR and it's working fine for my setup.

dustinbyrne:
The only required header, as far as I can tell, is the `editor-version`. I've prepended headers with `x-appmap-` and added a header for the Copilot plugin version.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
dustinbyrne: The requirement was that the `user-agent` would somehow identify AppMap/Navie. The CLI now sets and controls the `user-agent`.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -68,15 +71,22 @@ public boolean isCopilotAuthenticated() {
...
+        var baseHeaders = new HashMap<>(
+            Map.of(
                 "editor-version", getCopilotEditorVersion(),
-                "user-agent", GitHubCopilot.USER_AGENT,
                 "x-github-api-version", GitHubCopilot.GITHUB_API_VERSION,
-                "vscode-machineid", machineId,
-                "vscode-sessionid", UUID.randomUUID().toString()
+                "x-appmap-machineid", machineId,
+                "x-appmap-copilot-language-server-version", GitHubCopilot.LANGUAGE_SERVER_VERSION,
+                "x-appmap-copilot-plugin-version", CopilotAppMapEnvProvider.getCopilotPlugin().getVersion(),
```
dividedmind, line 80: I would keep the header names the official plugin uses where they match instead of prefixing them with `X-Appmap`; so `vscode-machineid`, `copilot-language-server-version`, `editor-plugin-version`. Let copilot also get the metadata.

plugin-copilot/src/main/java/appland/copilotChat/NavieCopilotChatRequestHandler.java
```diff
@@ -230,11 +234,16 @@ public void end() {
...
                 copilotModel,
                 asCopilotChatMessages(openAIRequest.messages()),
+                collectProxiedRequestHeaders(fullHttpRequest),
                 openAIRequest.temperature(),
                 openAIRequest.topP(),
                 openAIRequest.n());
     }
 
+    private @NotNull Map<String, String> collectProxiedRequestHeaders(@NotNull FullHttpRequest fullHttpRequest) {
+        return fullHttpRequest.headers().entries().stream().collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
dividedmind, line 244: Does it make sense to filter the headers? I'm slightly concerned about leaking transactional headers like `content-type`, `accept`, `content-size`, `if-updated-since` etc.

dividedmind (approved):

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
dustinbyrne: I've restored these.

dividedmind:
We're not going forward with this, right?

dividedmind:
@dustinbyrne @kgilpin do we want to revive this to pass on the user agent? Or will we want to add some extra data to it?

kgilpin:
User Agent is worth sending; it should include useful version info. A comment says that the CLI is controlling the User Agent but the CLI doesn’t know the plugin version. Maybe worth merging as is or maybe needs to add plugin version? It’s up to you guys.

PR #855, opened 2025-02-06
Download content exclusion lists from GitHub

dividedmind:
This pull request introduces several changes to enhance the Copilot plugin, including new functionality for handling content exclusions and updates to existing services and notifications.

### New Functionality:
* [`plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java`](diffhunk://#diff-f2e48992aa593951e22d2a1f51f9d70b2cc556d203efab514ae59f085dfafb08R1-R77): Added a new class to fetch and manage content exclusion rules from the Copilot API.
* [`plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilot.java`](diffhunk://#diff-b2fd844a2f86c4b4539119b7150dba414faea536ec4246a18a0df6bd1639b03eR20-R21): Introduced a new constant `INTERNAL_API_URL` for the Copilot internal API.

### Service Updates:
* [`plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java`](diffhunk://#diff-5f5ef13a11b808355167a89e3b25e2600b08bcb61fc5e2217904c463cc7e631bR56-R124): Added methods to download and ensure content exclusions are up-to-date. Modified the `createChatSession` method to handle content exclusion downloads. [[1]](diffhunk://#diff-5f5ef13a11b808355167a89e3b25e2600b08bcb61fc5e2217904c463cc7e631bR56-R124) [[2]](diffhunk://#diff-5f5ef13a11b808355167a89e3b25e2600b08bcb61fc5e2217904c463cc7e631bL117-R177)

### Notification Enhancements:
* [`plugin-core/src/main/java/appland/notifications/AppMapNotifications.java`](diffhunk://#diff-7473336df14fc297e0fe900fcb8dff631f06d4981fe9f97a3123e35d6020c087R320-R341): Added a new notification method `showCopilotExclusionDownloadFailed` to inform users when content exclusion downloads fail.
* [`plugin-core/src/main/resources/messages/appland.properties`](diffhunk://#diff-8423749c5b5ea32c141cf9d4cdaa0163471934f02cd341cf3eb7a9902e93e10eR295-R297): Added new notification messages for content exclusion download failures.

### Minor Updates:
* `.run/intellij-plugin [runIde].run.xml`: Updated script parameters to run with a newer IDE version by default. ([.run/intellij-plugin [runIde].run.xmlL7-R7](diffhunk://#diff-05f911c196589d75db7e6cd94e42c17e6fe72be9c8db1cb2c024dba717ab65c5L7-R7), [.run/intellij-plugin [runIde].run.xmlR21](diffhunk://#diff-05f911c196589d75db7e6cd94e42c17e6fe72be9c8db1cb2c024dba717ab65c5R21))
* [`plugin-copilot/src/main/java/appland/copilotChat/CopilotStartupNotificationActivity.java`](diffhunk://#diff-4691286119508929523e2ed132862962c9271f31595390b8473a84dcb1f0c4c3L30-R31): Refactored code to use a local variable for the Copilot service and added a call to ensure content exclusions are downloaded. [[1]](diffhunk://#diff-4691286119508929523e2ed132862962c9271f31595390b8473a84dcb1f0c4c3L30-R31) [[2]](diffhunk://#diff-4691286119508929523e2ed132862962c9271f31595390b8473a84dcb1f0c4c3R44-R49)
* [`plugin-copilot/src/main/java/appland/copilotChat/copilot/UpdatingCopilotToken.java`](diffhunk://#diff-7b24c3bc78f357518e6395df47fa7d4bc1f6015667f701add0589a25253ceb09L19-R19): Updated the token request URL to use the new `INTERNAL_API_URL` constant.

### DIFF
--- a/plugin-copilot/src/main/java/appland/copilotChat/CopilotStartupNotificationActivity.java
+++ b/plugin-copilot/src/main/java/appland/copilotChat/CopilotStartupNotificationActivity.java
@@ -27,7 +27,8 @@ public class CopilotStartupNotificationActivity implements StartupActivity, Dumb
         }
 
         // if the Copilot integration is enabled, but not authenticated, show a notification
-        if (!GitHubCopilotService.getInstance().isCopilotAuthenticated()) {
+        GitHubCopilotService copilotService = GitHubCopilotService.getInstance();
+        if (!copilotService.isCopilotAuthenticated()) {
             ApplicationManager.getApplication().invokeLater(AppMapNotifications::showCopilotAuthenticationRequired);
             return;
         }
@@ -39,6 +40,12 @@ public class CopilotStartupNotificationActivity implements StartupActivity, Dumb
                 AppMapApplicationSettingsService.getInstance().setCopilotIntegrationDetected(true);
                 ApplicationManager.getApplication().invokeLater(AppMapNotifications::showFirstCopilotIntegrationEnabled);
             }
+        }
+
+        try {
+            copilotService.ensureContentExclusionsDownloaded();
+        } catch (Exception e) {
+            // this is ok for now, it will be retried later
         }
     }
 }
--- a/plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java
+++ b/plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java
@@ -1,0 +1,77 @@
+package appland.copilotChat.copilot;
+
+import appland.utils.GsonUtils;
+import com.google.gson.annotations.SerializedName;
+import com.intellij.util.Url;
+import com.intellij.util.Urls;
+import com.intellij.util.io.HttpRequests;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+public record CopilotContentExclusion(@SerializedName("rules") @NotNull List<Rule> rules,
+                                      @SerializedName("last_updated_at") @NotNull String lastUpdatedAt,
+                                      @SerializedName("scope") @NotNull Scope scope) {
+    /**
+     * Fetches the global exclusion rules from the Copilot API.
+     *
+     * @param gitHubToken The GitHub token to use for authentication.
+     * @return The content exclusion rules.
+     * @throws IOException
+     */
+    static public @NotNull List<CopilotContentExclusion> fetchGlobal(@NotNull String gitHubToken) throws IOException {
+        return fetch(gitHubToken, Scope.ALL, null);
+    }
+
+    /**
+     * Fetches the content exclusion rules from the Copilot API.
+     *
+     * @param gitHubToken The GitHub token to use for authentication.
+     * @param scope       The scope of the content exclusion rules to fetch.
+     * @param repos       The specific repositories to fetch the content exclusion rules for
+     * @return The content exclusion rules.
+     * @throws IOException
+     */
+    static public @NotNull List<CopilotContentExclusion> fetch(@NotNull String gitHubToken, @NotNull Scope scope, @Nullable List<String> repos) throws IOException {
+        try {
+            var response = HttpRequests.request(contentExclusionUrl(scope, repos)).tuner(connection -> {
+                connection.setRequestProperty("Authorization", "Bearer " + gitHubToken);
+                connection.setRequestProperty("Accept", "application/json");
+            }).isReadResponseOnError(true).readString();
+            return List.of(GsonUtils.GSON.fromJson(response, CopilotContentExclusion[].class));
+        } catch (HttpRequests.HttpStatusException e) {
+            if (e.getStatusCode() == 404) {
+                return List.of();
+            }
+            throw e;
+        }
+    }
+
+
+    /**
+     * @param scope The scope of the content exclusion rules to fetch.
+     * @param repos The specific repositories to fetch the content exclusion rules for
+     */
+    static private @NotNull Url contentExclusionUrl(@NotNull Scope scope, @Nullable List<String> repos) {
+        var url = Urls.parse(GitHubCopilot.INTERNAL_API_URL + "/content_exclusion", false).addParameters(Map.of("scope", scope.name().toLowerCase()));
+        if (repos != null) {
+            return url.addParameters(Map.of("repos", String.join(",", repos)));
+        }
+
+        return url;
+    }
+
+    public enum Scope {
+        @SerializedName("repo") REPO, @SerializedName("all") ALL,
+    }
+
+    public record Rule(@SerializedName("source") @NotNull Source source,
+                       @SerializedName("paths") @NotNull List<String> paths) {
+    }
+
+    public record Source(@SerializedName("name") @NotNull String name, @SerializedName("type") @NotNull String type) {
+    }
+}
--- a/plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
+++ b/plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
@@ -1,8 +1,8 @@
 package appland.copilotChat.copilot;
 
+import appland.notifications.AppMapNotifications;
 import appland.utils.GsonUtils;
 import appland.utils.SystemProperties;
-import com.esotericsoftware.kryo.kryo5.util.Null;
 import com.google.gson.JsonObject;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Map;
 import java.util.UUID;
 
@@ -53,12 +54,71 @@ public final class GitHubCopilotService {
     }
 
     /**
+     * Whether the content exclusions have been downloaded.
+     */
+    private boolean contentExclusionsDownloaded = false;
+
+    /**
+     * Downloads the content exclusions from the GitHub Copilot API and writes them to the global ignores file.
+     */
+    private synchronized void downloadContentExclusions(@NotNull String githubToken) throws IOException {
+        if (contentExclusionsDownloaded) return;
+        var exclusions = CopilotContentExclusion.fetchGlobal(githubToken);
+        var paths = new ArrayList<String>();
+        for (var exclusion : exclusions) {
+            LOG.debug("Downloaded content exclusion: " + exclusion);
+            for (var rule : exclusion.rules())
+                paths.addAll(rule.paths());
+        }
+        var path = contentExclusionPath();
+        Files.createDirectories(path.getParent());
+        Files.write(path, paths);
+        contentExclusionsDownloaded = true;
+    }
+
+    /**
+     * Ensures that the content exclusions have been downloaded.
+     */
+    public void ensureContentExclusionsDownloaded() throws IOException {
+        if (contentExclusionsDownloaded) return;
+        var githubToken = loadGitHubOAuthToken();
+        if (githubToken != null) ensureContentExclusionsDownloaded(githubToken);
+    }
+
+    /**
+     * Ensures that the content exclusions have been downloaded.
+     */
+    private void ensureContentExclusionsDownloaded(@NotNull String githubToken) throws IOException {
+        try {
+            downloadContentExclusions(githubToken);
+        } catch (IOException e) {
+            LOG.warn("Failed to download content exclusions", e);
+            throw e;
+        }
+    }
+
+    /**
+     * @return The path to the global ignores file.
+     */
+    private static @NotNull Path contentExclusionPath() {
+        return Path.of(SystemProperties.getUserHome(), ".appmap", "navie", "global-ignore");
+    }
+
+    /**
      * Creates a new chat session with the GitHub Copilot API.
      * If the OAuth token or the Copilot token could not be retrieved, this method will return {@code null}.
      */
     public @Nullable CopilotChatSession createChatSession() {
         var githubToken = loadGitHubOAuthToken();
+
         if (githubToken == null) {
+            return null;
+        }
+
+        try {
+            ensureContentExclusionsDownloaded(githubToken);
+        } catch (IOException e) {
+            ApplicationManager.getApplication().invokeLater(() -> AppMapNotifications.showCopilotExclusionDownloadFailed(e));
             return null;
         }
 
@@ -114,7 +174,7 @@ public final class GitHubCopilotService {
     /**
      * @return The {@link Path} to the GitHub Copilot plugin's configuration file, or {@code null} if it could not be found.
      */
-    private static @Null Path findGitHubCopilotConfig() {
+    private static @Nullable Path findGitHubCopilotConfig() {
         var basePath = findCopilotSettingsDirectory();
         if (!Files.isDirectory(basePath)) {
             return null;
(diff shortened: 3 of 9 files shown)


gradle.properties
```diff
@@ -9,16 +9,16 @@ ideVersionVerifier=IC-2023.1,IC-2024.3,IC-251.17181.16
 lombokVersion=1.18.32
 
 # alternative versions to simplify development and testing
-ideVersion=2023.1
```
jansorg, line 12: This must not be committed, because it defines the version to build a release build and we have to build against the earliest supported version.
Changing it locally only is fine, of course.

gradle.properties
```diff
@@ -9,16 +9,16 @@ ideVersionVerifier=IC-2023.1,IC-2024.3,IC-251.17181.16
...
+#ideVersion=2023.1
 #ideVersion=2023.2
 #ideVersion=2023.3.2
 #ideVersion=2024.1
 #ideVersion=2024.2
-#ideVersion=2024.3.1
+ideVersion=2024.3.1
```
jansorg, line 17: Same as above...

gradle.properties
```diff
@@ -9,16 +9,16 @@ ideVersionVerifier=IC-2023.1,IC-2024.3,IC-251.17181.16
...
 #ideVersion=251.17181.16
 
-copilotPluginVersion=1.5.30-231
-#copilotPluginVersion=1.5.30-242
+#copilotPluginVersion=1.5.30-231
+copilotPluginVersion=1.5.30-242
```
jansorg, line 21: Same as above

plugin-copilot/src/main/java/appland/copilotChat/CopilotStartupNotificationActivity.java
```diff
@@ -27,7 +27,8 @@ public void runActivity(@NotNull Project project) {
         }
 
         // if the Copilot integration is enabled, but not authenticated, show a notification
-        if (!GitHubCopilotService.getInstance().isCopilotAuthenticated()) {
+        GitHubCopilotService instance = GitHubCopilotService.getInstance();
```
jansorg, line 30: nit: perhaps name this "copilotService" or similar?

jansorg (commented):
First pass, I'll post a more comments later

plugin-copilot/src/main/java/appland/copilotChat/CopilotStartupNotificationActivity.java
```diff
@@ -40,5 +41,9 @@ public void runActivity(@NotNull Project project) {
                 ApplicationManager.getApplication().invokeLater(AppMapNotifications::showFirstCopilotIntegrationEnabled);
             }
         }
+
+        // Try to download content exclusions.
+        // Note if it fails it will be retried later.
+        instance.downloadContentExclusions();
```
jansorg, line 47: A `StartupActivity` is invoked when a project is opened.
If 3 projects were restored at startup of the IDE, then the exclusions would be downloaded 3 times.
I'd name this method `ensureDownloadContentExclusions()` and make it it call `ensureContentExclusionsDownloaded(githubToken);` internally to only download the exclusions once.
`GitHubCopilotService` is an appliation-service, i.e. caching the state is fine.

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java
```diff
@@ -0,0 +1,75 @@
...
+                                      @SerializedName("last_updated_at") String lastUpdatedAt,
+                                      @SerializedName("scope") Scope scope) {
+    /**
+     * Fetches the global exclusion rules from the Copilot API.
+     *
+     * @param gitHubToken The GitHub token to use for authentication.
+     * @return The content exclusion rules.
+     * @throws IOException
+     */
+    static public List<CopilotContentExclusion> fetchGlobal(String gitHubToken) throws IOException {
```
jansorg, line 24: nit: I usually add `@NonNull` to make it clear what's expected.
```suggestion
    static public List<CopilotContentExclusion> fetchGlobal(@NotNull String gitHubToken) throws IOException {
```

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Fetches the content exclusion rules from the Copilot API.
+     *
+     * @param gitHubToken The GitHub token to use for authentication.
+     * @param scope       The scope of the content exclusion rules to fetch.
+     * @param repos       The specific repositories to fetch the content exclusion rules for
+     * @return The content exclusion rules.
+     * @throws IOException
+     */
+    static public List<CopilotContentExclusion> fetch(String gitHubToken, Scope scope, @Nullable List<String> repos) throws IOException {
```
jansorg, line 37:
```suggestion
    private static public @NotNull List<CopilotContentExclusion> fetch(@NotNull String gitHubToken, @NotNull Scope scope, @Nullable List<String> repos) throws IOException {
```

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java
```diff
@@ -0,0 +1,75 @@
...
+            throw e;
+        }
+    }
+
+
+    /**
+     * @param scope The scope of the content exclusion rules to fetch.
+     * @param repos The specific repositories to fetch the content exclusion rules for
+     */
+    static private Url contentExclusionUrl(Scope scope, @Nullable List<String> repos) {
```
jansorg, line 57:
```suggestion
    private static @NotNull Url contentExclusionUrl(Scope scope, @Nullable List<String> repos) {
```

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java
```diff
@@ -0,0 +1,75 @@
...
+        var url = Urls.parse("https://api.github.com/copilot_internal/content_exclusion", false).addParameters(Map.of("scope", scope.name().toLowerCase()));
```
jansorg, line 58: I'm wondering if we should extract the base URL into a `static final String`. We're already using it at `UpdatingCopilotToken`, too.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -52,13 +54,75 @@ public boolean isCopilotAuthenticated() {
         return registry.getEncoding(encodingType);
     }
 
+    /**
+     * Whether the content exclusions have been downloaded.
+     */
+    private boolean contentExclusionsDownloaded = false;
```
jansorg, line 60: I'd make this a `volatile` because it could be updated from different threads.
```suggestion
    private volatile boolean contentExclusionsDownloaded = false;
```

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -52,13 +54,75 @@ public boolean isCopilotAuthenticated() {
...
+
     /**
      * Creates a new chat session with the GitHub Copilot API.
      * If the OAuth token or the Copilot token could not be retrieved, this method will return {@code null}.
      */
     public @Nullable CopilotChatSession createChatSession() {
         var githubToken = loadGitHubOAuthToken();
-        if (githubToken == null) {
+
+        if (githubToken == null || !ensureContentExclusionsDownloaded(githubToken)) {
```
jansorg, line 125: This code is called from several places.
It's also invoked by `appland.copilotChat.ChooseCopilotModelAction`. I'm not sure if we really should make loading&choosing the model fail if the exclusions fail to launch.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -52,13 +54,75 @@ public boolean isCopilotAuthenticated() {
...
+     *
+     * @return Whether the content exclusions have been downloaded.
+     */
+    private boolean ensureContentExclusionsDownloaded(String githubToken) {
+        if (contentExclusionsDownloaded) return true;
+        try {
+            downloadContentExclusions(githubToken);
+            return true;
+        } catch (IOException e) {
+            ApplicationManager.getApplication().invokeLater(() -> AppMapNotifications.showCopilotExclusionDownloadFailed(e));
```
jansorg, line 106: Logging the exception would be helpful. There's no guarantee that a notification is actually shown to the user. It's possible to configure notifications to be always hidden.

jansorg: Currently, this notification would  be shown as part of "Choose Copilot model" if the download of the exclusions failed.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -52,13 +54,75 @@ public boolean isCopilotAuthenticated() {
...
+     */
+    private void downloadContentExclusions(String githubToken) throws IOException {
+        var exclusions = CopilotContentExclusion.fetchGlobal(githubToken);
+        var paths = new ArrayList<String>();
+        for (var exclusion : exclusions) {
+            LOG.debug("Downloaded content exclusion: " + exclusion);
+            for (var rule : exclusion.rules())
+                paths.addAll(rule.paths());
+        }
+        var path = contentExclusionPath();
```
jansorg, line 73: Do you want to write the file even if the list of exclusions is empty?

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
```diff
@@ -52,13 +54,75 @@ public boolean isCopilotAuthenticated() {
...
+        Files.createDirectories(path.getParent());
```
jansorg, line 74: The `Files` methods could throw exceptions, perhaps add a try/catch block?

jansorg (changes requested):
Thank you for the PR!
I added a few notes and thoughts.
I tested this and it's working for me (I think :smile:, it's a 404 because I'm not in an organization with exclusions, I suppose).

plugin-copilot/src/main/java/appland/copilotChat/CopilotStartupNotificationActivity.java
dividedmind: I don't think it should be called `ensure` since it's okay-ish if it fails. I'll add a check there to make sure it doesn't redownload it.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
dividedmind: This isn't ideal, but I think it's acceptable. Most of the time it will work; and if it doesn't, completion is not supposed to work anyway. @kgilpin WDYT?

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
dividedmind: Wouldn't it be better to make the download methods `synchronized`?

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
dividedmind: Yes — this means no exclusions. If the file is not present, the CLI currently falls back on a hardcoded list (which is not what we want).

plugin-copilot/src/main/java/appland/copilotChat/copilot/CopilotContentExclusion.java
dividedmind: Good idea.

gradle.properties
dividedmind: I modified the run configuration instead, is that OK? It's just a bit annoying to have to modify the properties to run locally with the newer IDE. I always forget and have to go back, flip it, and restart. If I recently cleaned my caches it also adds waiting for the downloads, twice.

plugin-copilot/src/main/java/appland/copilotChat/CopilotStartupNotificationActivity.java
dividedmind: Ok, I reworked it a bit, let me know if you think it's better.

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
jansorg: synchronizing the method would work without `volatile`, but doing this can potentially block for a longer time.
If possible, I'd like to avoid this.
With only `volatile` it would be possible that the download happens multiple times.

See https://medium.com/@alxkm/java-double-check-locking-pattern-3f8782e9d01e  for a helpful explanation how to fix this in Java.

Something like this should work:

```java

    /**
     * Whether the content exclusions have been downloaded.
     */
    private volatile boolean contentExclusionsDownloaded = false;

    /**
     * Downloads the content exclusions from the GitHub Copilot API and writes them to the global ignores file.
     */
    private synchronized void downloadContentExclusions(@NotNull String githubToken) throws IOException {
        if (!contentExclusionsDownloaded) {
            synchronized (this) {
                if (!contentExclusionsDownloaded) {
                    var exclusions = CopilotContentExclusion.fetchGlobal(githubToken);
                    var paths = new ArrayList<String>();
                    for (var exclusion : exclusions) {
                        LOG.debug("Downloaded content exclusion: " + exclusion);
                        for (var rule : exclusion.rules())
                            paths.addAll(rule.paths());
                    }
                    var path = contentExclusionPath();
                    Files.createDirectories(path.getParent());
                    Files.write(path, paths);

                    contentExclusionsDownloaded = true;
                }
            }
        }
    }
```

jansorg (changes requested):
Thanks!
I've suggested one change for the synchronized download.  My suggestion to only use `volatile` wasn't good enough. 
Otherwise this PR looks fine to me!

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
dividedmind: IIUC, this won't actually work as you intend - if downloadContentExclusions is already synchronized, the second synchronized block won't do anything except acquire another lock on the already locked object. Plus, volatile isn't needed since it's only accessed under a synchronization lock. I think you meant to not make the method synchronized.

Anyway, I don't think it's worth it to complicate this over just making the method synchronized (like I already did). It's not like it's a hot path with a lot of contention — the method will only be called concurrently a couple times at most at the start, and thereafter should only be called non-concurrently once per request (so human-scale time, seconds or even minutes apart) — and either way it will return instantly; and there's nothing else taking this lock. WDYT?

jansorg: Oh, of course the method must not be `synchronized`. I forgot to remove this.
I've tested a long download with a breakpoint (set to "suspend only thread"). It's working okay. Such a download blocks a background thread, but other project can still be opened (although they trigger the download again).

I'd probably move `contentExclusionsDownloaded = true;` into `private void ensureContentExclusionsDownloaded(@NotNull String githubToken)` because it has the exception handling and has `ensure` in it's name.

jansorg (approved):
I added a note, but it's up to you if you make that change.
Thanks for the PR!

plugin-copilot/src/main/java/appland/copilotChat/copilot/GitHubCopilotService.java
dividedmind: Thanks for the testing!

Actually, I'd rather keep the flag change with the code that changes the state — this ensures that it's always updated iff it succeeds; if it was separated then it'd be easier to overlook and/or make a mistake when refactoring later.

Issue #848, opened 2025-01-29
Navie sometimes won't initialize until the welcome message has completed


dustinbyrne:
This appears to be IntelliJ specific, but at the time of writing I am not sure of the root cause, so I'm logging it here.

The Navie frontend is blocked from loading while a welcome completion in another tab is pending. I _believe_ this is possible while a response completion is pending/generating as well.

One oddity of the video below is it does appear to load in parallel once, but once the tab is moved the page doesn't load. It's possible there was no pending completion (yet) when the second tab was initially loaded. This behavior doesn't need to be in side-by-side editor windows to occur, it's just easier to visualize in this manner.

https://github.com/user-attachments/assets/a03fbd99-8327-43ec-a3e2-2c1d8d0bbca7

jansorg:
current status: I'm unable to reproduce this on Linux

PR #857, opened 2025-02-10
Don't delay Navie UI until IDE finished indexing

jansorg:
Related to https://github.com/getappmap/appmap-intellij-plugin/issues/848

Previously, the Navie UI was only shown when the IDE finished indexing.
Especially at startup or when libraries are indexed, this can cause a large delay between opening Navie and the time the Navie UI is displayed.

This PR fixes this problem by detecting dumb mode. Properties are refreshed after the IDE turned into smart mode.

### DIFF
--- a/plugin-core/src/main/java/appland/webviews/navie/NavieEditor.java
+++ b/plugin-core/src/main/java/appland/webviews/navie/NavieEditor.java
@@ -103,9 +103,20 @@ public class NavieEditor extends WebviewEditor<Void> {
         var port = NavieEditorProvider.KEY_INDEXER_RPC_PORT.get(file);
         assert port != null;
 
-        var updatableNavieData = DumbService.getInstance(project).runReadActionInSmartMode(() -> {
-            return createUpdatableNavieData(project);
-        });
+        // If the IDE is in dumb mode (e.g. indexing files), then showing the Navie UI must not be delayed until
+        // smart mode is available.
+        // Instead, we pass empty data to Navie and trigger a refresh when the IDE switches to smart mode.
+        UpdatableNavieData updatableNavieData;
+        if (DumbService.isDumb(project)) {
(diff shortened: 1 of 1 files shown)


dividedmind (commented):
Looks good, but is there a potential race here? What if indexing finishes between setupInitMessage() and afterInit()?

dividedmind:
> @dividedmind You're right, this could happen. I've made a change to fix it.

I'm not sure I understand how the change fixes it. Is the following sequence possible?

- `DumbService.isDumb(project)` -> `updatableNavieData = new UpdatableNavieData(false, List.of());`
- `DumbService.exitDumbMode()` (subscription not set up yet, data not updated)
- `afterInit()` -> `busConnection.subscribe()` (it's too late to catch the transition above so the data will never be updated)

jansorg:
> > @dividedmind You're right, this could happen. I've made a change to fix it.
> 
> I'm not sure I understand how the change fixes it. Is the following sequence possible?
> 
> * `DumbService.isDumb(project)` -> `updatableNavieData = new UpdatableNavieData(false, List.of());`
> * `DumbService.exitDumbMode()` (subscription not set up yet, data not updated)
> * `afterInit()` -> `busConnection.subscribe()` (it's too late to catch the transition above so the data will never be updated)

I was too quick when I read your first comment and fixed another possible race, but not the one you mentioned.

I started to change the code to install the listener before the data is retrieved.
But this is problematic because it could trigger an update while the panel is still being initialized, i.e. it could interfere with the init. A better fix would be to trigger `updateNaviePropertiesAlarm` after the init finished, but this would introduce a delay due to the debouncing and Navie would not have up-to-date data when it's displayed for the first time.

I'd ignore this potential race because the solutions I can think of would make it worse.
WDYT?

dividedmind (approved):
Yeah, you're probably right, this is too much of an edge case to worry about and spend much effort. Just please add a comment about this for posterity.

Issue #767, opened 2024-08-08
Navie views contain large and distracting scroll bars


Daniel-Warner-X:
### Description

The scroll bars in the Navie response pane, context window, and code block panes appear very large and distracting compared to the other scrollbars in the JetBrains UI and the scrollbars as they render in Visual Studio Code. These scroll bars are very bright, wide, and numerous, which creates a distracting and unfinished experience when navigating Navie responses.

### Expected Behavior

The scroll bars in Navie views should match the appearance and behavior of scroll bars in the rest of the JetBrains Specifically:
- Scroll bars should be less bright and more subtle.
- Scroll bars should be narrower.
- The overall prominence and visibility of scroll bars should be reduced to avoid distraction.

### Actual Behavior

- Scroll bars in Navie views are very bright and wide.
- The scroll bars appear numerous and stand out compared to the rest of the IDE.
- Changing themes in the IDE does not affect the appearance of the scroll bars in Navie views.

### Impact

The large and distracting scroll bars give the experience of navigating Navie responses an unfinished feeling, making it uncomfortable and distracting for users.

### Steps to Reproduce

1. Open JetBrains and navigate to a Navie chat.
2. View the response pane, context window, and code block panes within Navie.
3. Observe the appearance and size of the scroll bars in these panes.
4. Compare these scroll bars to those in the rest of the JetBrains UI and Visual Studio Code.

### Environment Details

JetBrains: .72.0
Visual Studio Code: 0.125

### Additional Information

<img width="1499" alt="image" src="https://github.com/user-attachments/assets/457e9078-b4f3-41ed-b7ba-9d7503eab002">

jansorg:
With 2023.2 it's possible to style the scrollbar, https://youtrack.jetbrains.com/issue/IJPL-120103/JCEF-customize-scrollbars-in-JCEF,
but we can't use this yet because the plugin must be compatible with 2023.1 and later.

JetBrains is using https://kingsora.github.io/OverlayScrollbars/, it we want to keep compatibility with 2023.1, then we could perhaps  include this library and css instead of using CSS provided by the IDE?
But using the styling provided by the IDE would be preferred as we would not have to maintain our own solution.

jansorg:
On Slack it was discussed to put this on hold, for now.

jansorg:
This issue is related to https://github.com/getappmap/appmap-intellij-plugin/issues/764

Issue #764, opened 2024-08-08
Navie text size and style should match other UI elements more closely


Daniel-Warner-X:
**Description:**

The text in Navie's responses appears disproportionately large and cartoony compared to other UI elements within the IntelliJ IDE. This issue is noticeable across all of Navie's responses and does not respect the 'font size' setting that manipulates text in code files. Additionally, the text size in the VS Code version fits better with the overall default text size, providing a more consistent visual experience.

**Steps to Reproduce:**

1. Open IntelliJ IDEA with the AppMap plugin activated.
2. Trigger a response from Navie by asking a question or requesting help within the code editor.
3. Observe the text size and style of Navie's responses compared to other UI text (e.g., text in the file list).

**Expected Behavior:**

The text size and style of Navie's responses should be consistent with other UI elements in the IDE and respect the 'font size' setting for text in code files.

**Actual Behavior:**

- Navie's response text appears larger and cartoon-like compared to other UI text.
- The response text does not adjust according to the 'font size' settings that affect text in code files.

**Environment:**

- **IntelliJ IDEA Version:** 2024.1.4 by JetBrains s.r.o.
- **AppMap Plugin Version:** [insert version]
- **Operating System:** [insert OS and version]

**Visual Comparison:**

- **IntelliJ IDEA:** Navie's response text is larger and does not respect font size settings.

<img width="1613" alt="image" src="https://github.com/user-attachments/assets/22ed2f32-867d-4172-9279-b64b818593e0">

<img width="879" alt="image" src="https://github.com/user-attachments/assets/1c328c66-7963-43a9-a283-ce1a8239d8d6">

- **Visual Studio Code:** Navie's response text aligns more consistently with the default text size and overall UI.

<img width="1297" alt="image" src="https://github.com/user-attachments/assets/5205d65d-991f-4dfb-bd3a-3d9aaf2de864">

PR #862, opened 2025-02-18
Style AppMap webviews to match the IDE's theme

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/767
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/764

This PR provides CSS variables with colors and font properties to the AppMap webviews.
It also applies the IDE's scrollbar CSS style to the AppMap webviews, webview scrollbars should be more similar to the regular scrollbars of the IDE now.

Questions / open issues:
- [ ] Is there a property to provide the monospace editor font, e.g. to apply it to code snippets?
- [ ] I'm not sure how to style the white AppMap icon in the right panel ("Sequence diagrams")
- [ ] This needs testing on a Mac with a HiDPI screen to make sure the IDE's scale factor is properly applied

image


dustinbyrne:
It looks like the AppMap icon in the right doesn't adhere to the theme. We can fix this in a future release of the frontend, unless you want to shim it in via something like:
```css
.context__body__table-row__header svg path {
  fill: var(--appmap-color-foreground); 
}
```

It looks like the monospace font family is simply set to `monospace` - I'm not sure if JetBrains resolves that in some way or it's solely resolved by the OS.

dustinbyrne (approved):

PR #863, opened 2025-02-18
feat: Add logging for the AppMap RPC server

dividedmind:
This pull request introduces logging functionality to the `DefaultAppLandJsonRpcService` class in the `plugin-core` module. The changes include the creation of a log file for the JSON-RPC server and methods to handle logging of output and process termination events.

The log file can be found in the IntelliJ log directory:

https://github.com/user-attachments/assets/c1aba567-ed33-4abf-b467-f5577fbb2613

Key changes include:

### Logging Functionality:

* Added imports for `File`, `FileNotFoundException`, `FileOutputStream`, and `StandardCharsets` to support file operations and character encoding.
* Introduced a `FileOutputStream` instance `OUTPUT_LOG_STREAM` to manage log file output.
* Implemented `createOutputLogStream` method to create and initialize the log file stream.
* Added `logOutput` method to write log messages to the log file.
* Updated `onTextAvailable` and `processTerminated` methods to log relevant events using the `logOutput` method. [[1]](diffhunk://#diff-e4f5c05cbb0973f4c76566259c0c2924c79139141d82c069737443891c0eeaf1R441-R442) [[2]](diffhunk://#diff-e4f5c05cbb0973f4c76566259c0c2924c79139141d82c069737443891c0eeaf1R459)

### DIFF
--- a/plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
+++ b/plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
@@ -15,6 +15,7 @@ import com.intellij.execution.process.ProcessEvent;
 import com.intellij.execution.process.ProcessOutputTypes;
 import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.application.PathManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.DumbService;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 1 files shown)


dividedmind:
> It'd be nice to prepend the PID to each log message, but this is definitely an improvement.

Won't there be only one RPC server running most of the time anyway?

jansorg:
@dividedmind I was too late for a review.
The log is definitely useful for debugging.

With the current implementation, the log file is never closed. This is a (small) resource leak and I'd prefer if this was fixed.

Because `DefaultAppLandJsonRpcService` is a project service (i.e. it could have several active instances in a JVM) it wouldn't be sufficient to just make the field `non-static`.
The proper fix would be an application service, the field `non-static` in that service, make the service implement `Disposable`  and to close the stream in `dispose()`.
Let me know if you'd like me to make that change.

dividedmind:
> @dividedmind I was too late for a review. The log is definitely useful for debugging.
> 
> With the current implementation, the log file is never closed. This is a (small) resource leak and I'd prefer if this was fixed.
> 
> Because `DefaultAppLandJsonRpcService` is a project service (i.e. it could have several active instances in a JVM) it wouldn't be sufficient to just make the field `non-static`. The proper fix would be an application service, the field `non-static` in that service, make the service implement `Disposable` and to close the stream in `dispose()`. Let me know if you'd like me to make that change.

Yes, please, if you don't mind.

Issue #874, opened 2025-04-01
Integrate the model selector frontend


dustinbyrne:
**1. Introduction**

This outlines the necessary backend changes within the JetBrains AppMap plugin to support the new frontend AI Model Selector component. The goal is to replicate the functionality recently implemented in the VS Code extension, allowing users to select and configure various AI models (like OpenAI, Anthropic, Copilot/VSCode LM, etc.) for use with the Navie AI features.

The core change involves shifting from a simple configuration (API key + Copilot toggle) to a more flexible system where the backend RPC process manages the available models and their configurations, driven by plugin settings and UI interactions.

**2. Core Concepts & Goals**

*   **Flexible Model Support:** Allow users to choose from multiple AI providers and models.
*   **Configuration via UI:** Enable users to configure model-specific settings (API keys, base URLs) through the Navie webview UI.
*   **Centralized Backend Logic:** The AppMap RPC process (Node.js backend) will be the source of truth for available models and the currently selected model.
*   **Secure Secret Management:** API keys and other sensitive data must be stored securely using IDE-specific mechanisms (like JetBrains' `PasswordSafe`).
*   **Configuration Propagation:** Plugin settings and secrets need to be correctly passed as environment variables to the RPC process.
*   **Backend Restart:** Changes to sensitive configurations (like API keys stored in secrets) require restarting the RPC process.

**3. Required Changes in JetBrains Plugin (Java)**

**3.1. Configuration Management**

*   **New Setting:** Introduce a new persistent plugin setting to store the ID of the currently selected model.
    *   **ID:** `appMap.selectedModel` (or a suitable Java equivalent like `appmap.selected.model.id`)
    *   **Type:** String
*   **Existing Setting (Environment Variables):** The existing mechanism for storing non-secret environment variables will be used for non-sensitive model configurations (e.g., `OPENAI_BASE_URL`).
*   **New Secret Storage:** Implement a mechanism to store *secret* environment variables required by different models.
    *   **Mechanism:** Use JetBrains' `PasswordSafe` or a similar secure storage API.
    *   **Structure:** Store a collection of key-value pairs. Keys will be environment variable names (e.g., `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`).
    * **Usage:** The secret storage of environment variables is to be merged with non-secret environment variables from source code or from the user's own settings when launching the RPC service.

**3.2. API Key Migration (If Applicable)**

[...]

PR #876, opened 2025-04-05
Implement Navie model selector support

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/874

This PR implements the new Navie model selector.

There's no key/value store for settings, settings are usually identified by a getter/setter.
- `appMap.selectedModel` is implemented as `appland.settings.AppMapApplicationSettings#selectedAppMapModel`
- normal model config settings are implemented as `appland.settings.AppMapApplicationSettings#modelConfig` as a map to be able to store arbitrary keys and values. `null` values are dropped from the settings.
- secure model config settings are implemented as `appland.settings.AppMapSecureApplicationSettings#getModelConfig`, internally as a map to store arbitrary keys and values. The secure settings are stored in the IDE's password safe.

API Key Migration is implemented by `appland.settings.AppMapSettingsMigrationActivity`. A migrated settings is set to `null` after the successful migration.

Command line arguments like `--navie-provider local` are set by `appland.cli.DefaultCommandLineService#createAppMapJsonRpcCommand`.
The environment variables derired from the model config settings are applied by the same method.

The logic to restart the JSON-RPC server after changes to the environment variables is implemented in `appland.rpcService.DefaultAppLandJsonRpcService#DefaultAppLandJsonRpcService`.

The new JSON-RPC requests `v1.navie.models.add` and `v1.navie.models.select` are sent by `appland.rpcService.DefaultAppLandJsonRpcService#updateModelConfig`. With the current implementation, `v1.navie.models.select` is sent at the initial launch of the server and also after a restart.

New webview messages `change-model-config` and `select-model` are handled by `appland.webviews.navie.NavieEditor`.

Action "Choose Copilot model", which allowed the user to choose the copilot model for Navie using the IDE's UI, was removed because the model selection is now handled by the Navie webview.

### Unclear

- The settings UI still has an item to configure the OpenAI key. I wasn't sure if it's supposed to be removed or not.


dividedmind:
It seems to work reasonably well, with two bugs:
- right after start and for a moment after configuration (eg. OpenAI key) is changed, the dropdown is not populated and the instructions are shown instead (even if copilot is active). After a while the list populates again.
- there seem to be duplicates in copilot model list — I'm not sure if they are different versions?
Przechwycenie obrazu ekranu_2025-04-09_12-44-14

dustinbyrne (approved):

dustinbyrne:
> * right after start and for a moment after configuration (eg. OpenAI key) is changed, the dropdown is not populated and the instructions are shown instead (even if copilot is active). After a while the list populates again.

This is a limitation of the frontend. With the thread log, we should be able to push model updates instead of polling briefly after startup.

> * there seem to be duplicates in copilot model list — I'm not sure if they are different versions?

I've fixed this. If a model has duplicates it'll now include the version in the name.

PR #900, opened 2025-07-01
Update Gradle build setup

jansorg:
Unfortunately, we have to use the current snapshot of gradle-intellij-platform-plugin until it has a new stable release. The plugin introduced a breaking change and we have to adapt to fix our build setup.

Also, this PR is adding a workaround for https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/1971

It removes the fixed version of the plugin verifier, because the latest available version is now showing the bugs we experienced anymore.

The change to automatically fetch the latest Copilot plugin is unrelated, but helpful.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -245,11 +245,11 @@ project(":") {
 
     dependencies {
         intellijPlatform {
-            pluginModule(implementation(project(":plugin-core")))
-            pluginModule(implementation(project(":plugin-gradle")))
-            pluginModule(implementation(project(":plugin-java")))
-            pluginModule(implementation(project(":plugin-maven")))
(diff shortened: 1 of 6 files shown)


dividedmind (approved):
Looks good, thanks! I've been trying to figure it out yesterday, but I gave up. It's annoying how they keep breaking stuff.

One thing I'm concerned about is that if we use a snapshot version we're exposed to future breakage (since we cannot pin to a specific commit IIUC). I wonder if it would be possible to use a release version, adding any necessary workarounds, so at least the build is stable. Or maybe we can just do it on their next release...

jansorg:
Thanks for the review!

I'd switch to the next stable version as soon as it's available. I don't think that there's a previous stable version which does all we need...

PR #905, opened 2025-08-19
Revert "chore: install JetBrains JDK"

dividedmind:
This reverts commit b1837960e68114a55ce8aa7c486fc039d9df87dc.

JetBrains JDK doesn't seem to work with setup-java at the moment:
```
Run actions/setup-java@v4
  with:
    distribution: jetbrains
    java-version: 21
    java-package: jdk
    check-latest: false
    server-id: github
    server-username: GITHUB_ACTOR
    server-password: GITHUB_TOKEN
    overwrite-settings: true
    job-status: success
    token: ***
  env:
    GITHUB_TOKEN: ***
    JAVA_VERSION: 21
Installed distributions
  Trying to resolve the latest version from remote
  Error: Could not find satisfied version for SemVer '21'. 
  Available versions: 11_0_10b1145.115 (11.0.10+1145.115)
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
             - name: Setup Java
               uses: actions/setup-java@v4
               with:
-                  distribution: jetbrains
+                  distribution: temurin
                   java-version: ${{ env.JAVA_VERSION }}
 
             - name: Setup Gradle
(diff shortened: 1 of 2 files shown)


jansorg:
Mostly for consistency. JetBrains IDEs should be executed with a JetBrains runtime and the tooling is not clear which runtime is used when. For example, an SDK usually bundles JetBrains runtime, but I didn't not investigate yet if the JetBrains Gradle plugin uses the bundled runtime or not.
This seems more like a bug of the setup-java action, but to unblock using another JDK should work.

PR #911, opened 2025-10-14
Attach ZIP to job artifacts

jansorg:
Gradle task `verifyPlugin` builds the plugin artifact, which is then uploaded as an artifact. The upload action is only capable to take a directory, that's why the plugin zip is extracted first.

Successful run with a zip attached: https://github.com/getappmap/appmap-intellij-plugin/actions/runs/18507374754

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,9 +92,12 @@ jobs:
               id: artifact
               run: |
                   cd "${{ github.workspace }}/build/distributions"
+
                   FILENAME="$(ls *.zip)"
                   unzip -d content "$FILENAME"
-                  echo "filename=${FILENAME/.zip/}" >> $GITHUB_OUTPUT          
+
+                  OUTPUT_FILENAME="${FILENAME/.zip/}-build${GITHUB_RUN_NUMBER}.${GITHUB_SHA::6}"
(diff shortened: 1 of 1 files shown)


dividedmind:
> > > Looks good to me. The only question: do we _always_ need it in artifacts or maybe job will benefit from some guardrail (branch prefix etc.)?
> > > Not a blocker though, especially if we need it urgently right away.
> > 
> > 
> > What's the harm in always doing this? It takes just a couple seconds and can be IMO enormously useful.
> 
> No harm at all, rather two minor concerns:
> 
>     1. space consumption (44Mb per run) -- tolerable but not 100% neglectable in case of intensive development
> 
>     2. potential circulation of multiple plugin packs bearing the same version number, but different content, not traceable back to the build which generated them.

I wouldn't be too worried about space — IMO it's mostly negligible and we're not paying for this space directly — it's ephemeral anyway, basically like a pointer to cache. I agree about the version number (even though it's unlikely to be an issue in practice) — might be a good idea to add a job id or something similar to the filename at least. Could you do that and then just go ahead and merge? Thanks!

hleb-rubanau:
Now zip name is enhanced with build run number and sha 
https://github.com/getappmap/appmap-intellij-plugin/actions/runs/18519049716

PR #919, opened 2025-11-06
feat: Use GitHub releases for CLI version checks

dividedmind:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/920

Updates the CLI tool download service to fetch the latest versions from the GitHub releases API instead of the npm registry. This avoids issues in environments where npm is blocked.

The implementation has been refactored to cache the release versions to avoid redundant API calls. A test has been added to verify the parsing of the GitHub API response.

### DIFF
--- a/plugin-core/src/test/java/appland/utils/IndexTestUtils.java
+++ b/plugin-core/src/test/java/appland/utils/IndexTestUtils.java
@@ -1,42 +1,22 @@
 package appland.utils;
 
-import com.intellij.openapi.application.ApplicationInfo;
 import com.intellij.openapi.application.ApplicationManager;
-import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
+import com.intellij.testFramework.IndexingTestUtil;
 import org.jetbrains.annotations.NotNull;
 
-@SuppressWarnings("UnstableApiUsage")
 public class IndexTestUtils {
+    @SuppressWarnings("UnstableApiUsage")
     public static void waitUntilIndexesAreReady(@NotNull Project project) {
-        // IndexingTestUtil.waitUntilIndexesAreReady(project);
-        // 2024.2 made indexing in tests async and added IndexingTestUtil.waitUntilIndexesAreReady(project)
-        // to allow tests to wait for indexing to finish.
-        // But because this class and method is unavailable in < 2024.1, we need to use reflection to call this method.
-
-        if (ApplicationInfo.getInstance().getBuild().getBaselineVersion() >= 241) {
-            // according to JetBrains Slack, waitUntilIndexesAreReady must not be called inside a WriteAction or ReadAction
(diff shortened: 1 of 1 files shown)


plugin-core/src/main/java/appland/github/GitHubHttpRequests.java
```diff
@@ -0,0 +1,33 @@
...
+
+        return connection -> {
+            // Accessing api.github.com via GitHub actions often returns a 403,
+            // we're using the token to get a higher request rate limit.
+            // The system property is set by our Gradle build setup.
+            // https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions
+            var apiToken = System.getProperty("appland.github_token");
+            if (StringUtil.isNotEmpty(apiToken)) {
+                connection.setRequestProperty("authorization", "Bearer " + apiToken);
+            }
```
dividedmind: What is the source of the github token supposed to be — the environment variable or the system property?

jansorg: Sorry, my mistake. I merged two separate pieces of code. System property should be best, but it‘s sourced from the GitHub token by our build setup.

dividedmind:
> > > @dividedmind I've pushed a few additional commits to this PR:
> > 
> > 
> > Thanks!
> > > ```
> > > * Only cache pairs of CliTool and version, not the large JsonArray of all available versions (old and new). The JSON on disk was about ~380kb on disk.
> > > ```
> > 
> > 
> > Probably not strictly necessary, but a good catch! I didn't realize it was that big.
> > > ```
> > > * In unit test mode, always add the `Authorization` header to requests to GitHub. Adding the token in the header increases the rate limit and thus avoids breakage on CI due to too many requests.
> > > ```
> > 
> > 
> > I'd still be more comfortable if the tests avoided doing real network calls for reliability, but this is good enough for now, especially seeing we've already been doing that for other tests.
> 
> At least some of the tests need the CLI to work, so we have to make at least one request to download the binary. 

That's a good point.

> We may have to make this as part of the build (as with the agent), but this could potentially slow down the build time a lot (>100mb downloaded by gradle + up-to-date checks).

Yeah, no, that's definitely not what we want :)

> > > CI is still failing, but AFAIK only because of the known bug of the Java agent. This failure should go away as soon as there's a new release of the agent available.
> > 
> > 
> > The bug only happens when a test fails in the first place, so there must be something else going on there.
> 
> possibly, but it‘s hidden by the exception coming from the agent. But I don‘t think it‘s anything caused by this PR.

I agree, there seems to have been a heisenbug in the CI for a while, possibly a race or timeout — might even be an appmap-js bug. I've released new appmap-java and rerun this build (fingers crossed!); maybe we'll learn more.

PR #924, opened 2025-11-12
Invoke deploy workflow after successful build on main branch

jansorg:
This skips executing tests and plugin verification twice and is hopefully a more manageable solution.
The verification on `main` is still executed if dispatched manually.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,7 @@ on:
     push:
         branches:
             - develop
+            - main
     schedule:
         - cron: "0 0 * * 0"
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,6 +1,11 @@
 name: Deploy
 on:
     workflow_dispatch:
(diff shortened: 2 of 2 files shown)


jansorg:
@dividedmind I've updated the PR to use the `workflow_run` event.

dividedmind (approved):
I like this approach more as the previous one would run the deploy workflow as basically an action inside the build workflow. Having it be a separate workflow run is IMO cleaner and potentially more secure — in the future it would enable eg. limiting credential or other sensitive access to just the deploy workflow.

I note this will still probably fail due to disk space when run as manual dispatch. I'm leaning toward removing the verification steps altogether (except maybe perhaps verifyPluginProjectConfiguration) — this is going to be run automatically on passed tests, and if a human manually launches it, they would assume responsibility to make sure the code is okay.

dividedmind:
Hi @jansorg, I added the changes I mentioned (and also made sure build runs on pushes to main), could you take a quick look and if it looks good, go ahead and merge? Thanks.

PR #930, opened 2025-12-14
ci: Automatically publish the plugin to JB Marketplace on deploy

dividedmind:
Note I've already added the `Marketplace Deployment` environment containing the token. The environment is configured such that it will require a manual approval.


jansorg (approved):
I'm unable to test this, but the changes seem fine to me!

PR #937, opened 2026-02-13
Enable Library Source Navigation for AppMap References

dividedmind:
## Summary

Enables navigation to source code in external libraries (JARs, Python packages, Ruby gems) when clicking AppMap references. Previously, `showSource` only searched project content roots, resulting in "File not found" errors for framework classes like Spring or JDK classes.

(This PR also includes a regression test for executable permission fixes.)

## Implementation

Introduced an **extension point architecture** to maintain separation between language-agnostic (`plugin-core`) and language-specific (`plugin-java`) concerns:

### Key Components

1. **Extension Point (`plugin-core`)**
   - New `AppMapFileLookup` interface with `appland.appmap.fileLookup` extension point
   - Extensions are consulted when standard project content search fails
   - Enables language-specific lookup strategies

2. **Java-Specific Lookup (`plugin-java`)**
   - Uses `JavaPsiFacade.findClass()` to resolve fully-qualified class names
   - Automatically finds source attachments or falls back to decompiled view
   - Handles inner classes and classpath resolution

3. **Generic Fallback Lookup (`plugin-core`)**
   - Uses `FilenameIndex` with `GlobalSearchScope.allScope(project)`
   - Filters by relative path suffix matching
   - Runs with `order="last"` after language-specific lookups
   - Supports non-compiled languages (Python, Ruby, JavaScript)

4. **Smart Line Handling**
   - For `.class` files: ignores AppMap line numbers (opens at top)
   - Prevents confusing jumps to wrong locations in decompiled code
   - Preserves accurate navigation for source files

## Test Coverage

- Extension mechanism verification
- Java library class lookup
- Generic library file lookup (non-Java)
- Lookup priority ordering
- `.class` file line number handling
- Ambiguous match handling
- Executable permission regression test

### DIFF
--- a/plugin-core/src/main/java/appland/files/AppMapFileLookup.java
+++ b/plugin-core/src/main/java/appland/files/AppMapFileLookup.java
@@ -1,0 +1,31 @@
+package appland.files;
+
+import com.intellij.openapi.extensions.ExtensionPointName;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.vfs.VirtualFile;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Extension point to locate files referenced in AppMaps, e.g. from external libraries.
+ */
+public interface AppMapFileLookup {
+    ExtensionPointName<AppMapFileLookup> EP_NAME = ExtensionPointName.create("appland.files.fileLookup");
+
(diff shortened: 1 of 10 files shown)


jansorg (changes requested):
Thanks! I like the extension points 😄 
I added a few notes, let me know if you need more details.

PR #940, opened 2026-03-23
Fix storing of OpenAI key

### LINKED ISSUE #922: OpenAI key is not stored in model config and is reset at startup
I noticed that the OpenAI key stored in our settings is still saved as `secureApplicationSettings.openAIKey`. But this key is reset at startup and then saved into the model config settings. As a result, the OpenAI key is shown empty in the settings.
IMHO we should change the settings UI to get/set the model config value and not the deprecated secure setting above.

Image

### LINKED ISSUE #939: OpenAI key configured in settings does not override OpenAI key set from the webview
The two values are currently stored in separate properties and only the key set from the webview is passed to the JSON-RPC service.

jansorg:
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/939
Closes https://github.com/getappmap/appmap-intellij-plugin/issues/922

While testing I notived that the AppMap code review editor displayed an error with a different OpenAI key than the value saved via the AppMap settings UI. This was caused by configuring the OpenAI key also in the AppMap webview UI.


dividedmind (approved):
Good catch, thanks!

PR #943, opened 2026-04-16
Webview: Update dependencies and simplify polyfills, shims and the build

dividedmind:
These changes have been prompted by security advisories about some dependencies. While all of them were AFAICT benign, I figured it's still a good practice and excuse to update some dependency versions.

While many were trivial, a couple turned out to be transitive dependencies of polyfills we were using. This caused me to turn my attention to the polyfills and the build process, and I noticed:
- We were manually setting `mainFields` option for esbuild instead of relying on `platform: 'browser'` setting which [does more extensive setup](https://esbuild.github.io/api/#platform) in addition to the `mainFields` (which was slightly misconfigured anyway) to ensure a correct browser build. One interesting example is that it predefines `process.env.NODE_ENV` which allows tree-shaking to eg. skip bundling development (or production) versions of dependencies. In any case, this allowed bundling the correct versions of some dependencies, removing some uses of the polyfills.
- We were using a [super-library](https://www.npmjs.com/package/node-libs-browser) which pulled many different polyfills, more than we actually required. Many of them were CommonJS which esbuild couldn't properly tree-shake.
- We were polyfilling `globalThis` which for a long time hasn't been required in browser engines.

This pull request fixes these issues by streamlining the build process and dropping unused polyfills, in addition to bumping a few versions. As a bonus the bundle is 7 MiB smaller.

I did some manual testing and all the webviews appear to work correctly, but more testing is definitely welcome. It's possible I missed eg. a dynamic require somewhere that could cause something to break.


jansorg:
Thanks for the PR and apologies for the late reply.
The changes seem fine to me, the reduced size of the package is great.
I also did some manual testing and couldn't find a problem with the updates webviews.

@dividedmind I think we should publish the pending publish. If you agree, could you either approve https://github.com/getappmap/appmap-intellij-plugin/actions/runs/24777465818 or let me know?
Thanks!

PR #946, opened 2026-04-25
Model configuration fixes

dividedmind:
I looked at the flow of user configuring a custom OpenAI key or endpoint through the UI and found that it didn't work reliably — the current navie tab froze and the user was forced to open a new tab. I tracked it down to two problems: port not always being preserved between restarts, and the configuration change causing more than one restart (theoretically up to four, but I've only seen two). While trying to fix these, I noticed that the RPC server service has grown to be a bit of a mess, so I refactored it to explicitly model the state machine to more easily see what's going on.

I've also noticed and fixed a couple other problems such as:
- setting an openai key disabled copilot integration unnecessarily — this was required before model config could have been pushed with rpc;
- environment changes didn't automatically trigger rpc restarts.

I suggest reviewing commit-by-commit; the commit messages provide more detail.

To test manually:
- open Navie with copilot enabled and OpenAI not configured,
- use gear icon to enter the OpenAI API key,
- notice that after a brief pause Navie becomes available again,
- the model selector should now list both copilot and OpenAI models,
- check the appmap-rpc logs and verify the server only restarted once.


plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
```diff
@@ -293,109 +333,120 @@ private void restartServerAsync() {
...
+            lastKnownPort = port;
+            state = ServerState.RUNNING;
+            crashRestartCount = 0;
+        }
+        // Push config before notifying listeners so the initial state is already set when they react.
+        updateServerSettings();
+        updateModelConfig();
+        if (!isDisposed && !project.isDisposed()) {
+            project.getMessageBus().syncPublisher(AppLandJsonRpcListener.TOPIC).serverStarted();
+        }
```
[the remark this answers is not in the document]
dividedmind: This is intentional, so the server is completely ready before clients try to interact with it. I haven't observed any concerning delays in practice.

plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
```diff
@@ -200,86 +257,69 @@ protected BaseOutputReader.Options readerOptions() {
...
+    private void restartServerSync() {
+        ApplicationManager.getApplication().assertIsNonDispatchThread();
 
-        try {
-            var response = sendJsonRpcMessage(serverUrl, "v1.navie.thread.query", params);
-            if (response == null) {
-                throw new IOException("Received empty response from JSON-RPC server");
+        // No try/finally needed: stopCurrentProcessSync swallows all exceptions internally,
+        // and serverStarted fires automatically from onPortAnnounced when the new process announces its port.
+        project.getMessageBus().syncPublisher(AppLandJsonRpcListener.TOPIC).beforeServerRestart();
```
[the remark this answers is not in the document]
dividedmind: I don't think that's a real concern. It's an unlikely corner case and I don't think the consequences are significant.

plugin-core/src/main/java/appland/rpcService/TestAppLandJsonRpcService.java
```diff
@@ -16,9 +16,9 @@ public TestAppLandJsonRpcService(@NotNull Project project) {
      */
     public static void killJsonRpcProcess(@NotNull Project project) {
         var service = (DefaultAppLandJsonRpcService) AppLandJsonRpcService.getInstance(project);
-        var server = service.jsonRpcServer;
-        if (server != null) {
-            server.processHandler.killProcess();
+        var process = service.currentProcess;
```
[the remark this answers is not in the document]
dividedmind: Unlikely to matter in practice, but ok.

plugin-core/src/main/java/appland/webviews/WebviewEditor.java
```diff
@@ -209,16 +209,16 @@ private void setupJCEF() {
...
-            ApplicationManager.getApplication().invokeLater(this::openDevTools);
-        }
-
         if (AppMapNotifications.isWebviewProxyWarningRequired()) {
             AppMapNotifications.showWebviewProxyBrokenWarning(project);
         }
     }
 
     private void initWebviewApplication() {
+        if (Registry.is("appmap.webview.open.dev.tools", false)) {
```
jansorg, line 218: Did you observe a problem with the previous setup?

plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
```diff
@@ -120,66 +125,114 @@ public void rootsChanged(@NotNull ModuleRootEvent event) {
...
-        startServer(null);
+        startServerInternal();
     }
 
-    private void startServer(@Nullable Integer port) {
-        if (isServerRunning()) {
-            LOG.debug("AppMap JSON-RPC server is already running.");
-            return;
+    @Override
+    public void stopServerAsync() {
```
jansorg: We could drop `stopServerAsync` from the service interface, if we like. It's now only used by the service implementation itself.

plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
```diff
@@ -200,86 +257,69 @@ protected BaseOutputReader.Options readerOptions() {
...
-                var jsonRpcServer = new JsonRpcServer(process);
-                process.addProcessListener(new JsonRpcProcessListener(jsonRpcServer), this);
-
-                this.jsonRpcServer = jsonRpcServer;
-
-                // launch process and our process listener
+                process.addProcessListener(processListener, this);
+                currentProcess = process;
+                state = ServerState.STARTING;
                 process.startNotify();
```
jansorg: I would not call this in the `synchronized` block. It's can potentially take a while (invoking listeners,  etc.) and as far as I can tell it does not need to be synchronized.

plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
```diff
@@ -293,109 +333,120 @@ private void restartServerAsync() {
...
-    private void restartServerSync() {
-        ApplicationManager.getApplication().assertIsNonDispatchThread();
+    private void onPortAnnounced(@NotNull ProcessHandler process, int port) {
+        synchronized (this) {
+            if (process != currentProcess || state != ServerState.STARTING) {
+                return;
+            }
+            lastKnownPort = port;
+            state = ServerState.RUNNING;
+            crashRestartCount = 0;
```
jansorg, line 339: If I'm not mistaken, this implements "restart after crash" differently.
Now, the count is reset when the port is announced. A (repeated) crash after the port announcement is not counted against the max restarts and the server would keep restarting. 
Previously, it was counted, AFAIK.

I'm not sure what the better implementation is. There are advantages of this implementation, e.g. doing restarts after the server kept running. Your call what you'd like to do.

jansorg (approved):
The changes seem fine to me, only a few minor notes.
It worked well when I tested it.

plugin-core/src/main/java/appland/webviews/WebviewEditor.java
dividedmind: Yes, it only opened for me about 30% of the time.

plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
dividedmind: Oh yeah, good point!

plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
dividedmind: Good point. This is unchanged from the previous implementation but I suppose it's a good excuse to move it.

plugin-core/src/main/java/appland/rpcService/DefaultAppLandJsonRpcService.java
dividedmind: Yes, this is intentional. Previous implementation was IMO broken because the crash counter never reset; this led to the situation that when the server crashed and successfully restarted, a subsequent unrelated crash (say, hours later) would have the restart unnecessarily delayed or even (if it exceeded the implicit crash counter) suppressed. Not to mention this could have interfered with the tests which didn't have access to reset the crash counter (although this could be exposed if needed), even if I can't tell for certain if any test problems I've seen when adding more crash tests were caused directly by this, and this concern could be fixed just by resetting on explicit (re)start.

When the server announces the port it's initialized, up and running, and we can be pretty sure if it were to crash due to eg. corrupted binary or a similar issue, it would have crashed by now. The alternative would be to set up a timeout to reset the crash counter after it has been successfully running for some time (maybe a minute or so). Wdyt?

PR #949, opened 2026-05-31
fix: load missing dependency map stylesheet in JetBrains webview

baiweicoder-jpg:
The Dependency Map is almost blank in the JetBrains plugin while the same AppMap renders correctly in VS Code.

The webview bundle already contains dist/appmap.css, but appmap.html does not load it explicitly. This change adds the missing stylesheet reference.

Tested with:
- IntelliJ IDEA 2026.1 platform
- Gradle task: clean buildPlugin -PplatformVersion=261
- Existing .appmap.json file that renders correctly in VS Code

### DIFF
--- a/appland-webview/appmap.html
+++ b/appland-webview/appmap.html
@@ -3,6 +3,7 @@
 <head>
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./dist/appmap.css"/>
     <link rel="stylesheet" href="ide-styles.css"/>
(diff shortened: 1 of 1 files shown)


dividedmind:
Thank you, good catch! We must have missed this when we refactored the webviews a while ago. I've applied the fix in 76a32aa2aeb1f4c5e42e945414fd2e690a965ab9 with similar fixes for other webviews (they seemed to work fine without the stylesheet links but maybe there would be some subtle differences).

dividedmind:
Note the fix will be released in 0.83.3 (currently under review by JetBrains).

PR #963, opened 2026-09-15
Show server-sent errors in the OTP sign in screen

dividedmind:
Integrates https://github.com/getappmap/appmap-js/pull/2396 to show sign-in errors. Also, bumps qs and js-yaml in the webview.


dividedmind:
Integrated in #959

